Issue #131, opened 2019-01-15
License when contributing to open source software


miguelmorin:
## Detailed description
 - Provide a detailed description of the change or addition you are proposing or query or issue you need help with:

I am submitting a few pull requests to the TensorFlow code base. I need a [Contributor License Agreement](https://opensource.google.com/docs/cla/#external-contributors) for that. Does the Turing own that contribution or do I? If the Turing, do we have a Contributor License Agreement with Google?

- Tag project members that can help addressing the issue

I talked about this with @KirstieJane.

## Add labels

- Please choose the appropriate label(s) from our existing label list to ensure that your issue is properly categorized. This will help us to better understand and address your issue.

I am unable to add labels, maybe because I don't have the relevant rights.

KirstieJane:
Thank you @miguelmorin! I just pinged a tweet out and posted in slack - I'll keep this issue updated with the answer.

I'll also try to draft out a couple of sentences that can go in one of the chapters. Maybe Open Science? Maybe collaborating on github?

dingaaling:
Hello from 2022 @miguelmorin ! Hope all is well with you, many months later. The TTW Core team is following up on issues in our backlog and wanted to check if we now should close this issue? If so, feel free to comment or 👍  this comment. If not, please leave a comment on how we could revive this thread and how we can support this work.

We'll leave the issue open for now, and if we don't hear from you by the end of October we'll mark it as closed - but we are always happy to reopen it! Thank you!

miguelmorin:
Hi Jennifer! Yes, you can close this issue, as Martin's reply made clear that I owned the contribution and could sign the Contributer License Agreement myself. It's up to you if you want to include that in "The Turing Way".

dingaaling:
Thanks @miguelmorin!

PR #134, opened 2019-01-16
Address label assignment

pherterich:
As pointed out in #131 - not everyone will be able to add labels so I adjusted the label section to cover that going forward

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -12,5 +12,5 @@ We recommend doing this using
 ---
 __Update__
 
-## Add labels
-Please choose the appropriate label(s) from our existing label list to ensure that your issue is properly categorized. This will help us to better understand and address your issue.
+## Labels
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -12,5 +12,5 @@ We recommend doing this using
 ---
 __Update__
 
-## Add labels
-Please choose the appropriate label(s) from our existing label list to ensure that your issue is properly categorized. This will help us to better understand and address your issue.
+## Labels
+Not everyone will have contributing rights to add labels. Thus someone from the core project team will choose the appropriate label(s) from our existing label list to ensure that your issue is properly categorized and will be addressed by a suitable team member.
```
alexmorley: I don't think we need "and will be ... member." Maybe the issue will be addressed by someone whose not a team member anyhow :smile:

pherterich: Good point, I was waffling too much. Taken out now.

pherterich:
How do I let GitHub know that I addressed the changes?

PR #144, opened 2019-01-21
Update issue and pull request templates

KirstieJane:
I think we could make better use of the fact that html comments aren't rendered.

This pull request contains a new pull request tempate and includes my suggestions for re-formatting the issue template we have.

This is very much open for discussion! Thank you to everyone who's been using the current template, it really helps nudge folks to add the information we need 🤗

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -27,7 +27,8 @@ Please provide as much context as possible and link to related issues and/or pul
 - [ ] *Lorem ipsum dolor sit amet, consectetur adipiscing.*
 
 
-### Who can help
+### Who can help?
+
 <!-- We suggest using bullets (indicated by * or -) and filled checkboxes [x] here -->
 
 * *Lorem ipsum dolor sit amet, consectetur adipiscing.*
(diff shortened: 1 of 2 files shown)


pherterich:
I'm an idiot who should read the whole file, you have mentioned that - sorry. Will everyone know what "write access" means though?

KirstieJane:
Good question @pherterich - I think most people will know....but it’s always better to be explicit and not use unnecessary jargon!  I’m not sure of a better phrasing than “write access” though... any suggestions? Very happy to adjust!

pherterich:
Something like "If GitHub gives you the option to assign the appropriate label (or labels) to your issue, please do so. If you do not have the rights to add a label, the core team will do it for you." ?
It's not great yet, but maybe something you can work with @KirstieJane :smiley_cat:

KirstieJane:
Hey @pherterich - I've had a think about this phrasing, but I can't think of a way that doesn't reinforce a hierarchy. The problem is that you can only assign a label if you're added as a collaborator on the project.....and we might add people quite liberally, but the first few contributions will be through a pull request (not as an internal collaborator).

Would you mind keeping the text as it is at the moment and if lots of people are confused by that part, we'll either take it out or adjust it?

pherterich:
@KirstieJane happy to keep as is if we can't find a better phrasing. And I guess google can easily help to understand write access if somebody really wants to figure it out.

PR #182, opened 2019-02-01
[Quick fix] Rename file to be capatalised

r-j-arnold:
The chapter issue template was called "chapter_issue_template" but the template for the chapters is called "CHAPTER_TEMPLATE". Rename the chapter issue template to also have a capatilised name for consistency's sake.


r-j-arnold:
Thanks Kirstie

PR #194, opened 2019-02-12
Move prerequisites to start of chapter

r-j-arnold:
In the chapter template and in both chapters already in master move the chapter prerequisites to the start of the chapter. Previously they came after the summary and why the chapter was useful which can be a lot of text. There's no point wasting a reader's time introducing them to a topic they do not yet have the foundations to grasp.

- Whether they think this is a good idea

By submitting this pull request I confirm that:

- [x] all contributors to this pull request who wish to be included are named in [humans.md](humans.md).

### DIFF
--- a/chapters/reproducibility.md
+++ b/chapters/reproducibility.md
@@ -1,23 +1,23 @@
 # What is reproducible science?
 
+## Prerequisites / recommended skill level
+No previous knowledge needed.
+
 ## Summary
 There are several definitions of reproducibility in use.
 This chapter will lay out why reproducibility is important for science and scientists, provide an overview of the most common definitions and define reproducibility for the context of this handbook.
 
 ## How this will help you/ why this is useful
-This chapter sets out the definition of reproducibility that the Turing Way project team used when writing this handbook. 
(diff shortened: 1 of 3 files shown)


pherterich:
I personally would like to read a small paragraph of what this is about to get my interest and then if I'm interested and I need to read other chapters first, I would go to these. I think this is very much a personal preference thing and we can either do this as a vote within the team or leave for e.g. the workshops to get feedback and we can adjust when we do the final edit of the book.

r-j-arnold:
I can understand that. One of the reasons I'd like to move the prerequisite's up is because in writing the chapters I have/am I keep finding they interrupt the chapter flow where they are. Summary -> why this is useful -> in depth material feels more natural to me, having prerequisites in the middle feels like a side step where it is.

I'm happy to survey peoples opinions on the slack meeting today and see if there's a consensus one way or the other.

r-j-arnold:
Thanks @LouiseABowler! In that case could you go ahead and approve so I can merge?

r-j-arnold:
Nice catch @LouiseABowler, should be fixed now

PR #219, opened 2019-02-20
Add some notes on containerization

ots22:
Adds some notes for a talk on containerization that could be helpful for a TW chapter on containers


r-j-arnold:
Fantastic, thanks @ots22! Merging now

Issue #601, opened 2019-05-28
Add a script to check the latin abbreviations as part of CI


NatalieZelenka:
### Summary
Latin abbreviations, for example "e.g." are not very accessible. Screen readers will sometimes mispronounce them, pronouncing "e.g." as "egg" for example. For this reason they are against the style guide, but this can be easy to miss. 

Suggestion: adding a CI script to test this and let people know where they exist in the book (without causing it to fail).

### What needs to be done?
- [x] Write a script to check for latin abbreviations.
- [ ] Add the script to travis, so that it fails, but is allowed to fail (doesn't stop people from completing pull requests).

### Who can help?
Alex?

---

### Updates
None yet

PR #602, opened 2019-05-28
[WIP] Latin abbreviations

NatalieThurlby:
### Summary
Checks for latin abbreviations in the book.
Addresses (does not yet close) #601

- [x] Add script to check for abbreviations
- [ ] Included in travis (so that it fails, but that's allowed)

- [ ] Fails, but still allows merge.
- [ ] Everything looks ok?


NatalieThurlby:
Hi @alexmorley! Thanks for all your help on Tuesday... Clearly I have broken something if travis still hasn't ran :/. I don't think I will have time to mess around with this for a while, so should I take the latin checking script out of the travis file and have it print but not fail for now, so that it can be pulled and someone else could try to add in the travis if they want to... or do you have any other suggestions?

alexmorley:
@NatalieThurlby It's super weird... even if the build config (`.travis.yml`) was wrong it should have told you that there was a build error (which is different from a build failure - travis goes grey rather than red)...

Anyway I had a shot at why I understood from the docs. Not 100% though so :crossed_fingers: These things get quite intense when the repo has lot's of languages. For "normal" scientific repos it's much easier to understand.

alexmorley:
Looks like its working to me. @NatalieThurlby can you let me know if the comments/structure make sense. If they do we can merge else I can try add some more. (or we can link out to this https://docs.travis-ci.com/user/customizing-the-build#rows-that-are-allowed-to-fail)

image

Issue #616, opened 2019-05-28
Continuous integration script for internal links to glossary terms


cassgvp:
### Summary

Create continuous integration script to search for instances of glossary terms in the book content and cross reference to the glossary chapter and apply appropriate formatting. See also  #598 

### What needs to be done?

- [ ] More detail here.
- [ ] Write the code...!

### Who can help?

* *Lorem ipsum dolor sit amet, consectetur adipiscing.*
* *Lorem ipsum dolor sit amet, consectetur adipiscing.*

---

### Updates

*Lorem ipsum dolor sit amet, consectetur adipiscing.*

da5nsy:
This is an interesting idea. Did this ever go anywhere?

cassgvp:
I see the [Glossary ](https://the-turing-way.netlify.app/afterword/glossary.html) exists, but not sure how it was made??

da5nsy:
It appears to be written manually according to [these style guide instructions](https://the-turing-way.netlify.app/community-handbook/style/style-more-styling.html#glossary), using [this jupyter book standard](https://jupyterbook.org/en/stable/content/content-blocks.html#glossaries).

It looks like this formatting was introduced in #1624

Issue #598, opened 2019-05-28
Create Glossary chapter


cassgvp:
## Detailed description
This chapter will provide a single point for new contributors to access jargon free explanations of all the technical terms used in the book. The first version of the chapter may compile glossaries from existing chapters, and it should be continually updated as the book grows (maybe add updating glossary as an an action on the contributor guide?). 

Below is a dump of some thoughts I've collected so far:
- [x] Open a glossary chapter issue...!
- [ ] Use a tool such as The Up-Goer Six Text Editor (https://splasho.com/upgoer6/) to config that glossary descriptions are using most accessible language. 
- [ ] Consider how to distinguish internal glossary links from external page redirects. Formatting italics?
- [ ] See if there is a Jupyter book plugin for more formatting options, for example making use of reStructured text in admonition blocks 
- [ ] Create continuous integration script to search for instances of glossary terms in the book content and cross reference to the glossary chapter and apply appropriate formatting (see issue #616).

**Terms to make sure are included (collated from README.md and CONTIBUTING.md)**

- [ ] pull request
- [ ] ping
- [ ] git 
- [ ] version control
- [ ] Markdown
- [ ] Issue
- [ ] rendered
- [ ] Lorem ipsum
- [ ] bug
- [ ] Jupyter
- [ ] Binder, BinderHub
- [ ] review (in this context)

### Table of contents
Alphabetical list of glossary items

## Current status
This chapter is empty. If anyone would like to make a start they are more than welcome to do so.

 - [ ] Write chapter outline
 - [ ] Add material to the chapter
 - [ ] Combine materials into a readable chapter
 - [ ] Proofread
 - [ ] Request reviews
 - [ ] Address reviews
 - [ ] Merge to master branch.

## Updates
To avoid that others have to read through the full thread of comments, update the initial issue with important updates (e.g. decisions taken) regularly so that all the important information is available at one glance.

cassgvp:
I've started compiling this and will continue to work on it. Creating the issue as a place holder for now.

mjg211:
Shall we move current glossaries (for example the one at the end of Version Control chapter) over to this chapter?

cassgvp:
> Shall we move current glossaries (for example the one at the end of Version Control chapter) over to this chapter?

Makes perfect sense! Feel free to dump them in and I'll tidy.

mjg211:
Will do now 👍

cassgvp:
Going to create a chapter using the contributor guide.

[...]

PR #641, opened 2019-06-01
feat: Add basic glossary placeholder

masonlr:
### Summary

Addresses (but does not close): #598, #616

Add a centralised glossary.

* Adds a centralised glossary.
* Adds a proof-of-concept script for checking links to glossary items:
    1. We want to be able to link from the main text to glossary items.
    2. We want to know when we've used a term in the text but have forgotten to link to the glossary.

- [ ] Everything looks ok?

### DIFF
--- a/book/content/glossary/glossary.md
+++ b/book/content/glossary/glossary.md
@@ -1,0 +1,37 @@
+# Glossary
+
+#### Binder
+
+A web-based service which allows users to upload and share fully-functioning versions of their projects in an environment they define.
+
+#### Computational environment
+
+Features of a computer which can impact the behaviour of work done on it, such as its operating system, what software it has installed, and what versions of software packages are installed.
+
+#### Conda
+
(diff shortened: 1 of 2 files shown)


masonlr:
@rainsworth yes, please review. One question is how glossary terms should be structured. I've gone with, for example: 

```
#### Binder

 A web-based service which allows users to upload and share fully-functioning versions of their projects in an environment they define.
```

This way we can link to individual terms from the content markdown files.

rainsworth:
> @rainsworth yes, please review. One question is how glossary terms should be structured. I've gone with, for example:
> 
> ```
> #### Binder
> 
>  A web-based service which allows users to upload and share fully-functioning versions of their projects in an environment they define.
> ```
> 
> This way we can link to individual terms from the content markdown files.

If it's easy enough to link this way, then I'm happy with this. If others have strong opinions about it down the line, we can always revise!

PR #824, opened 2020-02-09
Chapter 8.2: Clarification on Conda-pip integration

anilbey:
### Summary

After having read the chapter I found it hard for the readers to understand how Conda is keeping track of the packages installed through `pip`. Also the current explanation is not always correct as one can install `pip` outside a Conda environment [1].  

References:
1. https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-pkgs.html#installing-non-conda-packages

* Clarifies under which condition conda tracks `pip` packages

- [x] The following people should be added to the [table of contributors](https://github.com/alan-turing-institute/the-turing-way/blob/master/README.md#contributors) in the README file: @anilbey

### DIFF
--- a/book/content/reproducible_environments/02/package-management.md
+++ b/book/content/reproducible_environments/02/package-management.md
@@ -156,7 +156,7 @@ pip install scipy
 ```
 
 will list 'scipy' package explicitely - as long as `pip` is installed inside the currently active conda environment.
-This allows Conda to keep track of the Python packages installed via `pip`.
+Unfortunately, when conda and pip are used together to create an environment, it can lead to a state that can be hard to reproduce. Specifically, running conda after pip may potentially overwrite or break packages installed via pip. One way to avoid this is by installing as many requirements as possible with conda, and then use pip. Detailed information can be read on the post [Using Pip in a Conda Environment](https://www.anaconda.com/using-pip-in-a-conda-environment/).
 
 Although Python packages have been used in many of the examples given here Conda packages do not have to be Python packages, for example here the R base language is installed along with the R package r-yaml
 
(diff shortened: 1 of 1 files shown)


book/content/reproducible_environments/02/package-management.md
```diff
@@ -155,7 +155,8 @@ This is the best way to install packages from within Conda as it will also insta
 pip install scipy
 ```
 
-still works.
+still works as long as `pip` is installed inside the currently active conda environment.
```
malvikasharan:
```suggestion
will list 'scipy' package explicitely - as long as `pip` is installed inside the currently active conda environment.
```

malvikasharan: I am still not sure in pip was mentioned for installing a package in the conda environment, or independently. Usually, it is not recommended to use pip inside conda as it leads to an environment that is not easily reproducible. May be it is worth mentioning that here (ref: https://www.anaconda.com/using-pip-in-a-conda-environment/).

book/content/reproducible_environments/02/package-management.md
```diff
@@ -155,7 +155,8 @@ This is the best way to install packages from within Conda as it will also insta
...
+This allows Conda to keep track of the Python packages installed via `pip`.
```
malvikasharan:
```suggestion
Unfortunately, when conda and pip are used together to create an environment, it can lead to a state that can be hard to reproduce. Specifically, running conda after pip may potentially overwrite or break packages installed via pip. One way to avoid this is by installing as many requirements as possible with conda, and then use pip. Detailed information can be read on the post [Using Pip in a Conda Environment](https://www.anaconda.com/using-pip-in-a-conda-environment/).
```

book/content/reproducible_environments/02/package-management.md
anilbey: But oftentimes the conda version of a package is outdated since the updates are released on pip and someone has to put extra effort to create the conda package. I agree it is worth mentioning the possible incompatibility issues that can arise by mixing pip and conda packages.

PR #1302, opened 2020-08-11
Style guide update: remove table for cross ref labels

malvikasharan:
### Summary

* When writing the style guide for reference, I had created a table to record all the labels. However, in practice it is not necessary due to the name convention we have defined for labels, which make the location of the label identifiable by name. Therefore, I am removing the table and updating the style guide.

* See that the style guide updated section is complete

- [ ] Everything looks ok?


malvikasharan:
@sgibson91 I am tagging you for action/CI issue here.

In this PR I am removing a markdown file. As per the current conditions in the script, it looks for the files that are updated. Which in this case is the file that is deleted, so of course it can't find the file and raises an error (failed CI). 😅 Any quick suggestion before I make any change in the test scripts to allow it to pass on this PR.

sgibson91:
Ah, so sorry @malvikasharan! I've opened a PR that I think should fix this (https://github.com/alan-turing-institute/the-turing-way/pull/1304) - it's not the most elegant solution but should work in a pinch 🤞

malvikasharan:
Thanks @sgibson91. The CI works with the new exception commands of ignoring the deleted files. 🎉

PR #1740, opened 2021-02-20
Indent numbered list blocks to avoid number reset

inwaves:
### Summary

Fixes small grammatical error regarding noun determiner.

### List of changes proposed in this PR (pull-request)

* On the Netlify-hosted _Turing Way_, the numbered list indices are reset because of the code blocks at each point. This means the list looks as below, with 1. as the index for all points. To fix this, indent the code blocks with 4 spaces.
image

### What should a reviewer concentrate their feedback on?

- [x] Ensure fix remedies the list indexing on Netlify.

### Acknowledging contributors

- [ ] All contributors to this pull request are already named in the [table of contributors](https://github.com/alan-turing-institute/the-turing-way/blob/master/README.md#contributors) in the README file.
- [x] The following people should be added to the [table of contributors](https://github.com/alan-turing-institute/the-turing-way/blob/master/README.md#contributors) in the README file: @inwaves

Issue #1670, opened 2020-12-02
[BUG] Fixing external links


martinagvilas:
### Summary

The way some external links are written trigger a warning in the Jupyter-Book build. External links should be written down with `https://` (e.g.  `https://twitter.com`) and not with `www.` (e.g.  `www.twitter.com`) or without prefix (e.g. `twitter.com`)

### What needs to be done?

Replace the external links written down with `www.` and without prefix, with external links written down with `https://`

The following links should be corrected:

- [x] link `twitter.com/` in line 23 of file `collaboration/new-community/new-community-guide.md`
- [x] link `arewemeetingyet.com` in line 121 of file `collaboration/remote-collab/remote-collab-conferences.md`
- [x] link `osf.io/meetings` in line 187 of file `collaboration/remote-collab/remote-collab-conferences.md`
- [ ] link  `www.slack.com` in line 29 of file `collaboration/remote-collab/remote-collab-participation.md`
- [ ] link `www.twitter.com` in line 29 of file `collaboration/remote-collab/remote-collab-participation.md` 
- [ ] link `www.github.com` in line 29 of file `collaboration/remote-collab/remote-collab-participation.md` 
- [ ] link `www.gitter.im` in line 29 of file `collaboration/remote-collab/remote-collab-participation.md` 
- [ ] link `www.grammarly.com` in line 19 of file `community-handbook/consistency/consistency-language.md`
- [ ] link `arewemeetingyet.com/` in line 89 of file `community-handbook/coworking/coworking-organisation.md`
- [ ] link `theturingway.slack.com` in line 38 of file `community-handbook/newsletters/newsletters-process.md`
- [ ] link `tinyletter.com/` in line 96 of file `community-handbook/newsletters/newsletters-process.md`
- [ ] link `theturingway.slack.com` in line 116 of file `community-handbook/newsletters/newsletters-process.md`

### Who can help?

Everyone!

malvikasharan:
@Hanafi can you please take on this issue?

Hanaffi:
@malvikasharan Yes, Can you assign it to me please?
Btw, Its `Hanaffi` not `Hanafi`  (Double 'n') 😁

malvikasharan:
Corrected your username. It's all yours.

malvikasharan:
Warning that still exist:
- File: afterword/contributors-record 
  - (line  557) broken    https://umontreal.openscience.ca
- collaboration/github/github-maintainers 
[...]

PR #1779, opened 2021-03-05
Add "tips" section

inwaves:
### Summary

Add a "tips" section containing to ensure moving forward #1740 and #1670 are avoided. 

* Added "tips" section to offer style guidance for items that triggered previous issues.
* Also fixed two typos in the text.

- [ ] Special content blocks look ok?

### DIFF
--- a/book/website/community-handbook/style.md
+++ b/book/website/community-handbook/style.md
@@ -71,10 +71,10 @@ Any chapter containing a Latin abbreviation will fail the continuous integration
 ## Tips
 
 ```{note}
-Indent the content following each item of a numbered list.
+Indent the content following each item of a numbered list. If the content is unindented, the list numbering will be reset.
 ```
 
-Please avoid:
+For instance, please avoid:
 1. First list item
 ```
 This content belongs to the first list item.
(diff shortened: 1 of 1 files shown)


EstherPlomp (commented):
Hi @inwaves! Is it correct that there is no incorrect example for the 2. second list item? 
Cheers, Esther

inwaves:
Hi @EstherPlomp 

Yes—basically what I wanted to show was that indenting the blockquote after the first list point resolves the issue with the reset numbering:
image

I will add a bit more context around why it's necessary, because right now it isn't clear what issue we're trying to avoid.

EstherPlomp:
Ah thanks, that makes more sense now! 
In the deploy preview the wrong numbering is displayed correctly, however, the notes are also kind of strange visually: 
image
Would it be possible to remove the note parts and have it displayed similar as the rest of the guide? 

I also think it makes sense to have the note on 'Write external links using http:// instead of the www.' as a separate section with a little explanation about why people should use http://? Cheers!

EstherPlomp (approved):
Thanks very much for adjusting, I think it is very clear now!

Issue #1955, opened 2021-05-23
[WIP] Updating the landing page


da5nsy:
### Summary

Sketching out some ideas for modifications to the landing page, based on discussions and ideas from the Community Share-out and Contributor Celebration on 21st May 2021.

- Assumption of new readers/contributors that one actually needed to read the entire guide, linearly.
- The realization that that is actually a bad idea now considering the size/scope.
- The idea of changing the landing page to better mirror how people actually interact with the guide: searching and/or looking through contents lists.

#### Proposal:
Additional line to the intro, along the lines of *"The guide has now grown to a size where it is frankly inadvisable to attempt to read the entire thing from start to end (and what even is the start and the end?) - most users will probably be better off starting by searching for a [topic of interest](link to search?) or browsing the [contents](link to summary?)".*

### What needs to be done?

- [ ] Add additional text (worth getting the wording spot on).

### Who can help?

* Anyone!

---

### Updates

- I've started a WIP PR (#1961) and added an initial commit (https://github.com/alan-turing-institute/the-turing-way/pull/1961/commits/da0476b03e86e0e36c3102b708692d82258f1a9e)
- I've added bullet points to that PR detailing next steps.

da5nsy:
Noting that this has some crossover with #1851 - to the extent that they both are to do with folks' initial impression of The Turing Way often being overwhelming.

da5nsy:
Addressed in #1961 (will probably be followed by further changes from #2035, #2033, and #2032.

Issue #1919, opened 2021-05-18
[WIP] Self-reflection prompts on ethical research practice


ninadicara:
## Summary

This chapter is a suggested contribution to the Guide for Ethical Research. Ethical dilemmas are complex, and whilst ethics checklists can help us to simplify decision making processes, we also need to understand the biases and world views that we approach these decisions with.

Self-reflection is a great tool for this! It supports us to critically examine our own thought processes, think about power and privilege in the context of a particular research project and be better informed about our own biases. Reflective skills can be applied to the way we work in teams, the way we work with the communities we research, or the research we choose to do or not do. 

Reflection can be difficult though, and wouldn't usually be something data scientists are trained in, unless as part of training to teach. This chapter proposes a series of guided prompts that could be helpful for anyone wanting a refresher or place to start thinking about how to apply self-reflection to their data science work. 

### Discussion Points

- [ ] Which prompts would be most helpful?
- [ ] Are the prompts accessible for everybody?
- [ ] Have we referenced the right thinkers in this area? What theory do we need to make sure we have mentioned?
- [ ] Have we incorporated prompts that cross a broad range of perspectives?
- [ ] Are these prompts helpful for those doing data science work with no human data? For instance, they could still apply to working in a team. 

## Who can help?

This chapter is aimed at anyone in any field of data science, so everyone will be able to help write it!

* People working on the Guide to Ethical Research for advice on where this chapter will fit.
* Anyone interested in helping to write self-reflection prompts.
* Anyone willing to try out the ones in the draft and give feedback! 

## Updates

- [ ] Write chapter outline
- [ ] Add material to the chapter
- [ ] Combine materials into a readable chapter
- [ ] Proofread
- [ ] Request reviews (inc. trialling the prompts)
- [ ] Address reviews
- [ ] Merge to master branch.

ninadicara:
Note - think this would have relevant cross-overs with other currently open issues such as #1315 and #1236.

Ismael-KG:
Super quick recap of suggestions following our chat yesterday, @ninadicara!

1. Maybe work off (or find inspiration in) a case study of a researcher who needs to reflect on their work (perhaps someone who is unreflectively using racial racial categories or in the realm of brain organisation theory)
2. Ismael's waffling
3. Maybe focus on the discovery of one's assumptions (these can be fed into by education; one's experiences relative to their ethnicity, culture, age, gender, (dis)ability, class... -- but focussing on these sources can be complicating)

[...]

Issue #1851, opened 2021-04-20
On-the-fly, personal curation of the book's content


sgibson91:
### Summary

Paraphrased from @KirstieJane:

We find that The Turing Way book is super intimidating for folks. There are loads of chapters, lots of content and I often hear "I'll read the book and get back to you" and .... well..... they don't. Because there's too much there to digest!

BUT - what if we could host subsections of the Jupyter Book in a "choose your own adventure" or "curate the reading list" format?

Then we could have views that said "Are you new to reproducibility? Would you like to learn about data management? How much time do you have (5 mins, 15 mins, 1 hour)?" and then someone could click through to answer those questions and see BOOM - their personally curated book!

All the content would be the same but we'd serve just a subsection of the content. I think it makes the book more accessible AND it really showcases the power of WHY you'd want to host a book online rather than just print out a PDF etc.

There's another small point which could fit in which is to customise the views - for example making a cover page - would be useful! We've found that seeing all the possibilities right at the start is what makes people anxious.

### What needs to be done?

- [ ] Scope the technical possibility
- [ ] *Add further items here*

### Who can help?

Skills required:

* Mapping of personas onto content
* Designing the initial questionnaire
* Backend processing of questionnaire responses into personas/content mapping
* Frontend development for accessing the questionnaire and serving the generated mapping

---

### Updates

*Lorem ipsum dolor sit amet, consectetur adipiscing.*

PR #1961, opened 2021-05-30
Updating the landing page

da5nsy:
### Summary

as per #1955, to make it less intimidating (#1851, #1919, probably others) and to direct readers to the recommended ways of interacting with the guide (browse or keyword search).

- I've tried to cut down the initial message to the bare minimum
  - I've moved out some of the text into a new 'history' section. **This would need expanding or removing before a merge** - it is currently a [stub](https://en.wikipedia.org/wiki/Wikipedia:Stub#Basic_information). I imagine this could be a good place for origin stories, key people, funding thanks etc. This interacts with #1731 - probably makes sense to include a link to a fuller history rather than have a duplicate 'history' on the landing page, unless the plan was for it to be there anyhow.
  - I've cut words where I could.
- I've tried to keep the language impersonal (avoiding language like 'we recommend...' because 'we' seems like an odd concept in this setting). This has meant that the language is quite direct (The option of 'We recommend not trying to read the book from start to end' has been passed over in favor of 'Do not try to read the guide from start to finish'). **Suggestions on how to make this sound friendlier are welcome!**
- I would like to be able to link to a summary page (really just the navigation pane but on a page of its own). I can't find a way to get jupyter book to create this page through URL manipulation, though there might be a way that I haven't found. **Help requested in finding this.**
- Talking or URL manipulation, I assume that there is a better way to link to the search page than with the full hardcoded URL. I hope that someone who understands how this works better than me will be able to **fix this**.
- (Unrelated to the main goal - I've hyphenated 'open source' since it is an adjective in this case.)

Bolded above, but to summarise:
~~- [ ] Extend/remove history section (see note above regards #1731)~~ The history section will arrive at some point, at which point we can work out the best way to link to it.
~~- [ ] Make wording friendlier~~ It's actually fine.
~~- [ ] Find a way to generate a full-page clone of the navigation page, to be linked to from 'browse the chapters'~~ I'm actually happy with people having to use the normal navigation; it's good practice to have them use the standard navigation rather than creating a new/different/redundant navigation path.
~~- [ ] Fix my hacky link to the search page~~ I've actually removed it, see above logic.

### Updates

- I decided that everything was actually fine! (See crossed out to do list)

### DIFF
--- a/book/website/welcome.md
+++ b/book/website/welcome.md
@@ -1,14 +1,11 @@
 (welcome)=
 # Welcome
 
-_The Turing Way_ is an open source community-driven guide to reproducible, ethical, inclusive and collaborative data science.
+_The Turing Way_ is an open-source community-driven guide to reproducible, ethical, inclusive and collaborative data science.
 
-Our goal is to provide all the information that data scientists in academia, industry, government and the third sector need at the start of their projects to ensure that they are easy to reproduce and reuse at the end.
+Our goal is to provide all the information that data scientists in academia, industry, government and the third sector need to ensure that the projects they work on are easy to reproduce and reuse.
 
-The book started as a guide for reproducibility, covering version control, testing, and continuous integration.
-However, technical skills are just one aspect of making data science research "open for all".
-
-In February 2020, _The Turing Way_ expanded to a series of books covering reproducible research, project design, communication, collaboration, and ethical research.
+Do not try to read the guide from start to finish. Take a [buffet approach](https://mobile.twitter.com/rossdavism/status/1138441421143400453) instead: browse the chapters or [search](https://the-turing-way.netlify.app/search.html) for something you would like to learn about.
 
 ```{figure} figures/welcome.jpg
 ---
@@ -60,6 +57,14 @@ Accordingly, everyone who participates in _The Turing Way_ project is expected t
 All contributions must abide by our [code of conduct](https://github.com/alan-turing-institute/the-turing-way/blob/master/CODE_OF_CONDUCT.md).
 
 ![Gif showing screen capture of contributors table, smiling faces and emojis representing the types of contributions in a table](https://media.giphy.com/media/gKIUisnjpj2PS75nOJ/giphy.gif)
+
+(welcome-history)=
+## History
+
+The book started as a guide for reproducibility, covering version control, testing, and continuous integration.
+However, technical skills are just one aspect of making data science research "open for all".
+
+In February 2020, _The Turing Way_ expanded to a series of books covering reproducible research, project design, communication, collaboration, and ethical research.
 
 (welcome-citing)=
 ## Citing _The Turing Way_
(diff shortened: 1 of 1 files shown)


da5nsy:
Just seen that I've unknowingly undone fae00bf0d0cfdb0604c277ab0fed52dc64b44a91 - I do think that "open-source" reads better (vs. "open source") in this context, but I'm curious to see if @paulowoicho or @malvikasharan have strong feelings?

- https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/o/open-source
- https://www.computerhope.com/jargon/o/opensour.htm

malvikasharan:
We use Open Science as a noun, that refers to a field of research/practice. Hyphenated open-science is an adjective, which means the same but is not used to refer to open science/open research/open source ...

So to avoid inconsistency with how open science is referred to in literature/community, I would keep it unhyphenated.
> Just seen that I've unknowingly undone [fae00bf](https://github.com/alan-turing-institute/the-turing-way/commit/fae00bf0d0cfdb0604c277ab0fed52dc64b44a91) - I do think that "open-source" reads better (vs. "open source") in this context, but I'm curious to see if @paulowoicho or @malvikasharan have strong feelings?

da5nsy:
Hi @malvikasharan 

I'm talking about the specific phrase "The Turing Way is an open-source community-driven book" where "open-source" is an adjective and "book" is the noun.

I feel quite strongly that it should be hyphenated in this situation.

ha0ye:
I see @da5nsy's point, but I see "open source" as an adjective very frequently, so the hyphen throws me off. Describing a book as "open source" also feels slightly odd; and I wonder if "community-developed community-driven" would work as an alternative.

da5nsy:
@ha0ye 
I like the idea of circumnavigating the problem, but I'm not sure "community-developed" is the way to go. For me, the important point that is communicated by the phrase "open-source" is that people can actually see the underlying code for the project.

To be honest, I was already a little unhappy with this sentence - it's the first sentence people read and it's super long and jumbly (however, we fall down on hyphenation). I'm going to change this PR back to WIP and try to improve it.

malvikasharan:
@ha0ye Using "Open Source" for books is not unusual (For [example](https://www.freecodecamp.org/news/taking-off-the-successful-launch-of-an-open-source-book-7553a2262898/)). The concept comes from open source framework for project development, which applies to different areas of open science. 

Another option is to use open book (let's not talk about hyphenation again 😅) project (as used by [O'Reilly](https://www.oreilly.com/openbook/)).
@da5nsy My suggestion is to not get stuck on details as we are going to launch a Governance proposal (see #2035), which will include updating our vision and mission statements that most members agree with. The vision statement will then become the opening line of the book.

da5nsy:
> The vision statement will then become the opening line of the book.

Aha, good to know. Thanks @malvikasharan. I shan't strain too much over getting it perfect then.

da5nsy:
I think this is ready for review now.

Main goals of this PR: 
- Clarify the landing page text
   - previously it was quite wordy, and "open source" (however you like to hyphenate it) was doing some heavy lifting. I've broken that out into "open source", "open contribution" and "open licenced" which I think better explain what we mean by "open source" (and do so in a way that is hopefully more understandable to folks who are new to the terms).
- Add a 'warning' not to try to read the whole thing.
   - I'm quite happy with this now, and I think this will be useful to newcomers.

These commits may be shortlived due to #2035, #2033, and #2032, but I think this works as a temporary modification, and there may well be pieces that could be incorporated long term.

malvikasharan (approved):
@da5nsy this looks really great! I like the breaking down of open to -source, -contribution, -license. 
Also, I absolutely adore the addition of the buffet approach. I am happy with how it looks and approve to merge it.

PR #2010, opened 2021-07-05
Remove advice to use C++ testing frameworks for C

rpep:
### Summary

Despite many people thinking that C++ is compatible with C, this is not true in the modern language as C99 diverged, introducing features that were not available in C++ and vice versa. This has continued, with C11 introducing further incompatabilities. This means that you can only universally use a C++ testing framework if you restrict yourself to the C89 standardwhich is obviously now 32 years old!

Because of that, using a C++ testing framework for a C project can run into issues. This is not obvious at first glance, but consider the following which my be present in a header file which would need to be included in a translation unit for a C++ testing framework to make use of a C library:

```
// 'restrict' is not a keyword in C++ so this generates a compiler error:
void myfunction(double* restrict array)

// C++ does not support this because it implements templates which achieves a similar thing:
#define myfunction(X) _Generic((X),
                   int: myfunctioni, \
                   long: myfunctionl, \
                   float: myfunctionf, \
                   double: myfunctiond, \
                   default: myfunction_default)(X)

// C++ doesn't support designated initialisers that can appear in a header:
const int a[] = {[0 ... 9] = 1, [10 ... 99] = 2, [100] = 3 };
```
Some modern C features can be enabled in some C++ compilers by using compiler flags, but this isn't universally true, which makes it hard to recommend. People used to working with GCC often run into issues when switching compilers, because GCC has many language extensions that for e.g. MSVC does not support.

* Remove "all C++ frameworks" from the list of suggested C testing frameworks

- [ ] Is removing the advice sensible, or should caveats be added?
- [ ] Everything looks ok?

- [x] The following people should be added to the [table of contributors](https://github.com/alan-turing-institute/the-turing-way/blob/master/README.md#contributors) in the README file: rpep

### DIFF
--- a/book/website/reproducible-research/testing/testing-guidance.md
+++ b/book/website/reproducible-research/testing/testing-guidance.md
@@ -63,7 +63,6 @@ Find one you like, learn about the features it offers, and make use of them. Com
   - Boost::Test
   - google-test
 - C
-  - all C++ frameworks
   - Check
   - CUnit
 - Python
--- a/book/website/reproducible-research/testing/testing-guidance.md
+++ b/book/website/reproducible-research/testing/testing-guidance.md
@@ -63,8 +63,12 @@ Find one you like, learn about the features it offers, and make use of them. Com
   - Boost::Test
   - google-test
 - C
+  - all C++ frameworks
   - Check
   - CUnit
+```{note}
+While modern C++ and C are still mostly compatible, they're not completely and using test framework interchangeably may not always work.
+```
 - Python
   - pytest (recommended)
   - unittest comes with standard Python library


book/website/reproducible-research/testing/testing-guidance.md
```diff
@@ -63,7 +63,6 @@ Find one you like, learn about the features it offers, and make use of them. Com
   - Boost::Test
   - google-test
 - C
-  - all C++ frameworks
```
c-martinez, line 66: Instead of removing this sentence, we could add a [note](https://the-turing-way.netlify.app/community-handbook/style/style-more-styling.html):

```{note}
While modern C++ and C are still mostly compatible, they're not completely and using test framework interchangeably may not always work.
```

c-martinez:
Hi @rpep,

I'm no C/C++ expert, but after consulting with a colleague, my understanding is that you are absolutely right. The only "issue" we could see with your proposed change is that it goes a bit too far. As far as I understand, C++ frameworks will work for C code a for the majority of cases, except when you run into corner cases as the one you describe.

I think it would be more useful for the reader to know that C++ frameworks work for C most of the time, but that they might run into issues in some cases. I would go for adding caveats as you already suggested.

I left a concrete suggestion on the relevant line.

I hope this is useful and thanks for your suggestion!

rpep:
@c-martinez sounds good to me - do you want me to make the change or do you want to apply your existing suggestion?

c-martinez:
@rpep if you could make the change please, that would be fantastic!

Thanks [@egpbos](https://github.com/egpbos) for your help behind the scenes!

c-martinez:
@rpep I had some time today to implement this change and see if it looks fine on my local machine. If you think it all looks fine, we can approve and merge it?

rpep (commented):
lgtm

rpep:
Hi @c-martinez - apologies for the delay - I was on holiday from work.

That looks fine to me!

Best,
Ryan

c-martinez (approved):

c-martinez:
Hi @rpep , no worries. Approved and merged.

Thanks for your help!

PR #2016, opened 2021-07-06
Tweak CI triggers to exclude all-contributors bot

sgibson91:
Duplicate of https://github.com/alan-turing-institute/the-turing-way/pull/2015

### Summary

Some PRs do not require CI tests to be run, such as those from the all-contributors bot. This PR changes the triggers of our tests to **exclude** branches that match `all-contributors/**` and **includes** pushes to paths that match the `book` directory. This should make **when** our tests run be more efficient. (See commit messages for more context.) The Netlify config file has also been updated to **not** deploy previews for branches that contain `all-contributors`.

Issue is that a lot of our CI tests are required for merging - do we lift these protection rules?

* Added conditional to exclude branch names containing "all-contributors"
* Added positive path pattern matching
* Added ignore pattern for `all-contributors` branches to Netlify `deploy-preview` context in config file (This might not take effect until merged?)

- [ ] Everything looks ok?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,10 +6,14 @@ on:
     branches:
       - master
       - "!all-contributors/**"
+    paths:
+      - "book/**"
   pull_request:
     branches:
       - master
       - "!all-contributors/**"
+    paths:
+      - "book/**"
 
 # set the default shell to bash for all jobs
 # this is important for the windows OS!
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,11 @@ on:
   push:
     branches:
       - master
+      - "!all-contributors/**"
   pull_request:
     branches:
       - master
+      - "!all-contributors/**"
 
 # set the default shell to bash for all jobs
 # this is important for the windows OS!
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,13 +5,11 @@ on:
   push:
     branches:
       - master
-      - "!all-contributors/**"
     paths:
       - "book/**"
   pull_request:
     branches:
       - master
-      - "!all-contributors/**"
     paths:
       - "book/**"
 
@@ -24,6 +22,7 @@ defaults:
 # run job
 jobs:
   build-jb-and-check-links:
+    if: (github.event.pull_request) && !contains(github.head_ref, 'all-contributors')
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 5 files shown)


malvikasharan:
@sgibson91 you're a star! Thank you for this PR <3 I will test this later today.

martinagvilas:
Looks awesome @sgibson91! 

Regarding the CI tests required for merging, in my opinion those could be lifted without too much harm if people are warned that they should not merge if tests are failing. But on the other hand, I think @malvikasharan mostly merges the PRs from all-contributors and I guess she has the power to merge without tests passing, right?

Also, one question for personal interest only: when you add `"!all-contributors/**"` to branches on push or pull request in the CI file, are you requesting it not to be triggered when a push or PR is done to this branch, or when a push or PR from this branch is made to another (or both?).

sgibson91:
> Also, one question for personal interest only: when you add `"!all-contributors/**"` to branches on push or pull request in the CI file, are you requesting it not to be triggered when a push or PR is done to this branch, or when a push or PR from this branch is made to another (or both?).

Ok so now I'm thinking about it

- adding `"!all-contributors/**"` to the `push` trigger is probably redundant, as push is only triggered on the main branch
- adding `"!all-contributors/**"` to `pull_request` trigger will prevent tests from running on PRs that want to merge **into** branches that match `all-contributors/**`

Hmmmm, this isn't as simple as I thought...!

How about this solution?

1. Revert the changes to the triggers
2. Add the following condition to the jobs

```yaml
if: (github.event.pull_request) && !contains(github.head_ref, "all-contributors")
```

Ref: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#contains

sgibson91:
> For a pull_request event, only branches and tags on the base are evaluated.

Ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

sgibson91:
> I think @malvikasharan mostly merges the PRs from all-contributors and I guess she has the power to merge without tests passing, right?

Unfortunately no, we have "Include administrators" checked so there is no override power.

martinagvilas:
> How about this solution?

Sounds like a good plan!

> Unfortunately no, we have "Include administrators" checked so there is no override power.

I see, then maybe the easiest way would be to allow to merge to main without the CI passing right?

sgibson91:
> Sounds like a good plan!

Fab I pushed that change 🎉 

> I see, then maybe the easiest way would be to allow to merge to main without the CI passing right?

I agree. @malvikasharan what do you think? PRs will still be protected by the required review rule, but not have CI passing required

malvikasharan:
@sgibson91 and @martinagvilas I trust your judgement on this. I like the idea of protected PR for review but skipping CI. 🙌

sgibson91:
@malvikasharan Great! I've lifted the CI requirement, feel free to merge/test whenever you have time :)

malvikasharan (approved):

Issue #2051, opened 2021-07-27
Update the translated README to the new version of the English README


BatoolMM:
### Summary

Thanks to the help of the wonderful @malvikasharan and @da5nsy, who just updated the README 🎉 They made it look **stunning** 😍 with more visual components and a better description of the project and the contribution process in PR #2032. However, we had the old README translated into **8 languages** which are **Korean, Dutch, French, Indonesian, German, Italian, Portagues and Spanish**. Therefore, we'd appreciate any help with getting them updated with the new version.
 
### What needs to be done?
- [ ] *🙋 Claim this issue: Comment below. If someone else has claimed it, ask if they've opened a pull request so maybe you can review it.*
- [ ] *Update the [Korean version of the README.](https://github.com/alan-turing-institute/the-turing-way/blob/master/README-translated/README-Korean.md)*
- [ ] *Update the [French version of the README.](https://github.com/alan-turing-institute/the-turing-way/blob/master/README-translated/README-French.md)*
- [ ] *Update the [Dutch version of the README.](https://github.com/alan-turing-institute/the-turing-way/blob/master/README-translated/README-Dutch.md)*
- [ ] *Update the [Indonesian version of the README.](https://github.com/alan-turing-institute/the-turing-way/blob/master/README-translated/README-Indonesian.md)*
- [ ] *Update the [German version of the README.](https://github.com/alan-turing-institute/the-turing-way/blob/master/README-translated/README-German.md)*
- [x] *Update the [Italian version of the README.](https://github.com/alan-turing-institute/the-turing-way/blob/master/README-translated/README-Italian.md)*
- [ ] *Update the [Portagues version of the README.](https://github.com/alan-turing-institute/the-turing-way/blob/master/README-translated/README-Portuguese.md)*
- [x] *Update the [Spanish version of the README.](https://github.com/alan-turing-institute/the-turing-way/blob/master/README-translated/README-Spanish.md)*

### Who can help?

* *This is a first-timers issue. This means we welcome contributions from individuals who either haven't contributed to our codebase before or even folks who haven't contributed to open source before..*
* *Anyone who translated the earlier versions of README.*
* *Anyone with knowledge of any of the earlier languages.*

---

perluna:
I'd like to give it a go to the Italian translation/update if nobody has claimed that yet :smiley:

BatoolMM:
Thank you @perluna, please let me know if you need any help with the PR. I'm looking forward to this Italian update :)

[...]

PR #2032, opened 2021-07-21
Harmonise README and Landing file 

malvikasharan:
### Summary

Partially address #2031 

* Updaing README to harmonise as discussed with @BatoolMM and @da5nsy 

### List of changes proposed in this PR (pull-request)

- [ ] Update logo
- [ ] Add links
- [ ] Add images

### What should a reviewer concentrate their feedback on?

- [ ] Everything looks ok?

### Acknowledging contributors

- [ ] All contributors to this pull request are already named in the [table of contributors](https://github.com/alan-turing-institute/the-turing-way/blob/master/README.md#contributors) in the README file.
- [ ] The following people should be added to the [table of contributors](https://github.com/alan-turing-institute/the-turing-way/blob/master/README.md#contributors) in the README file:

PR #2052, opened 2021-07-27
fix broken links in readme

da5nsy:
### Summary

A couple of broken links snuck through #2032. 

- [ ] Do the links now work?
- [ ] Everything looks ok?

### Future ideas

- It would be swell if we were allowed to put each new link on a new line to make the code easier to read, but markdown interprets that as a request for a new line of the table. If anyone knows a way around that, let me know! Googling only comes up with results for people who *want* new lines from single code lines (`<br>`), not anything for my case where I want a single output line from multiple code lines.
- It would be cool to pull in the contributors badge to the table, but it looks quite fragile (because [contributors bot needs to be able to find it to update it](https://allcontributors.org/docs/en/bot/faq#all-contributors-badge-count-does-not-update)) and to test whether it would break when moved into the table requires more than just seeing whether it renders, it would require asking the bot to add a new contributor to see whether it is able to automatically update it. 

### Update 
I've added the `review request` label to get a quick review on this so that it can be fixed before work gets underway on #2051


malvikasharan (approved):
Thank you! 🛠

PR #2159, opened 2021-11-08
Tidy up bibliography by removing empty doi field

da5nsy:
### Summary

Tidy up bibliography by removing empty doi field


Arielle-Bennett (approved):
Neat & tidy! 🧹

PR #2177, opened 2021-11-09
Create CITATION.cff file for The Turing Way

malvikasharan:
Add citation.cff file with metadata


malvikasharan:
@sgibson91 the previous release with zenodo metadata did not work properly with upload_type "publication". Looking at the error message, I am being prompted to create a Citation.cff file. For testing can you please check and approve it? I will then make sure that there are some documentation on the process for the future releases.

BatoolMM (approved):

PR #2404, opened 2022-05-20
try 1 git4research

jcolomb:
just a PR to test some issues with figures


TheAbstract (commented):
@jcolomb I think it's better to mark this as draft or WIP using GH indicators so that people and machines (bots) know how to process this. Also, if what you're done [testing] and don't intend on anything here being merged into main, then it's better to close the PR.

PR #2413, opened 2022-05-26
Refactor release_workflow.py to use a YAML parser

sgibson91:
CITATION.cff is a valid YAML file, so we use the ruamel YAML parser to
read/edit/write the file instead of manipulating individual lines.

Additionally added the workflow_dispatch trigger to the release-drafter.yml workflow file for testing

### DIFF
--- a/tests/release_workflow.py
+++ b/tests/release_workflow.py
@@ -7,9 +7,10 @@ from ruamel.yaml import YAML
 
 # Configure the YAML parser
 yaml = YAML()
-yaml.indent(mapping=3, sequence=2, offset=0)
+yaml.indent(mapping=2, sequence=4, offset=2)
 yaml.allow_duplicate_keys = True
 yaml.explicit_start = False
+yaml.preserve_quotes = True
 
 
 def update_citation_file(citation_file_path):
--- a/tests/release_workflow.py
+++ b/tests/release_workflow.py
@@ -17,6 +17,7 @@ def update_citation_file(citation_file_path):
         file_content = yaml.load(stream)
 
     major, minor, tiny = file_content["version"].split(".")
+    major, minor, tiny = int(major), int(minor), int(tiny)
 
     # Maybe we want to provide an input to determine what the release version
     # should be? Here we are restricting the patch and minor versions to be less
--- a/tests/release_workflow.py
+++ b/tests/release_workflow.py
@@ -1,46 +1,43 @@
 """
 Script to update citation.cff file and validate it then draft a release note.
 """
-import argparse
 from datetime import datetime
 
-def read_from_file(infile):
-    with open(infile, "r", encoding="utf8") as fid:
-        return fid.readlines()
+from ruamel.yaml import YAML
 
-def int_to_version(major, minor, tiny):
-    return f"{major}.{minor}.{tiny}"
+# Configure the YAML parser
+yaml = YAML()
+yaml.indent(mapping=3, sequence=2, offset=0)
+yaml.allow_duplicate_keys = True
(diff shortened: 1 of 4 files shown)


tests/release_workflow.py
```diff
@@ -1,46 +1,44 @@
...
+    if (tiny > 10) and (minor < 10):
+        tiny = 0
+        minor += 1
+    elif (tiny > 10) and (minor > 10):
+        minor = 0
+        major += 1
 
-            version_number_str = str(major) + "." + str(minor) + "." + str(tiny) + "\n"
+    version_str = f"{major}.{minor}.{tiny}"
+    file_content["version"] = version_str
```
sgibson91, line 34:
```suggestion
    file_content["version"] = f"{major}.{minor}.{tiny}"
```

tests/release_workflow.py
```diff
@@ -1,46 +1,43 @@
...
-            tiny = tiny + 1
-            if (tiny > 10 and minor < 10):
-                tiny = 0
-                minor = minor + 1
-            elif (tiny > 10 and minor > 10):
-                minor = 0
-                major = major + 1
+    # Maybe we want to provide an input to determine what the release version
+    # should be? Here we are restricting the patch and minor versions to be less
+    # than or equal to 10, but there's nothing in semver practices that enforces this.
```
timothy22000, line 24: Good idea! My current implementation is based on my discussions with Malvika. 

If we change this to take an input, how should we implement it as a variable input for the CLI execution in Github workflows since it will be dependent on how we classify changes in the TW?

timothy22000, line 24: Good idea! My current implementation is based on my discussions with Malvika. 

If we change this to take an input, how should we implement it as a variable input for the CLI execution in Github workflows since it will be dependent on how we classify changes in the TW?

tests/release_workflow.py
```diff
@@ -1,46 +1,43 @@
...
 from datetime import datetime
 
-def read_from_file(infile):
-    with open(infile, "r", encoding="utf8") as fid:
-        return fid.readlines()
+from ruamel.yaml import YAML
+
+# Configure the YAML parser
+yaml = YAML()
+yaml.indent(mapping=3, sequence=2, offset=0)
```
sgibson91, line 10:
```suggestion
yaml.indent(mapping=2, sequence=4, offset=2)
```

tests/release_workflow.py
```diff
@@ -1,46 +1,43 @@
...
+yaml.allow_duplicate_keys = True
+yaml.explicit_start = False
```
sgibson91, line 12:
```suggestion
yaml.explicit_start = False
yaml.preserve_quotes = True
```

tests/release_workflow.py
sgibson91: Although, I also noticed that release drafter is capable of providing a recommended version bump. So perhaps we can link those 2 actions and pipe the recommended version from release drafter into this script?

timothy22000: Interesting, I haven't heard about ChatOps before and have to read about it. Piping the recommended version from release-drafter into script sounds good. I presume that you are referring to the Version Resolver and the $NEXT_PATCH_VERSION variable for release drafter?

I was looking into semver and came across these two approaches that can automatically update the version number and adhere to semver guidelines as well based on the commit messages or tags.

The two approaches (one of them is a Github Action):

i) https://github.com/marketplace/actions/git-semantic-version
ii) https://github.com/semantic-release/semantic-release

Although we will still need to pipe into the script, so probably release-drafter will be sufficient. Just thought that I would share :)

PR #2570, opened 2022-08-11
Create give_a_talk.yml

aldenc:
new issue template for community members to record when they are giving a talk and get access to resources and a checklist

### Summary

new template for contributors to inform the community about an upcoming TTW presentation, with resources and checklist for preparing slides

### DIFF
--- a/.github/ISSUE_TEMPLATE/give_a_talk.yml
+++ b/.github/ISSUE_TEMPLATE/give_a_talk.yml
@@ -1,0 +1,43 @@
+name: Give a TTW talk 
+description: Resources and checklist for giving a talk about the Turing Way
+title: "[TALK] <title>"
+labels: [talks-and-workshops, newsletter]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        This template will guide you through the steps necessary to give a successful talk about The Turing Way!
+  - type: input
+    id: date
+    attributes:
+      label: Date of talk
+      description: When will you be presenting about the Turing Way?
+      placeholder: please use the format YYYY-MM-DD
+    validations:
+      required: true
+  - type: input
+    id: talk_details
+    attributes:
+      label: Details of the talk
+      description: Where are you presenting? Please provide a link to the event if available
+  - type: checkboxes
+    id: checklist
+    attributes:
+      label: Pre-talk checklist
+      description: Steps to prepare your presentation
+      options:
+        - label: Schedule your talk and let the community know in TTW Slack. If you’d like feedback and/or to schedule a practice talk, ask in the TTW Slack!
+        - label: Download template(s) from the [promotion pack](https://drive.google.com/drive/folders/1mzGmbJkPnP5q1goQesxDc_E5zAPL0eTF?usp=sharing) to ensure stylistic concistency
+        - label: If you re-used slides from a specific talk, please acknowledge the original author of the slides 
+        - label: Double check acknowledgements slide for TTW team, license info, Scriberia link, any additional acknowledgements
+        - label: Double check contact info for TTW links (book, Twitter, GitHub, Slack, newsletter), your own contact into
+  - type: checkboxes
+    id: checklist
+    attributes:
+      label: Post-talk checklist
+      description: Steps to share your presentation
+      options:
+        - label: Upload your slides as a PDF to [Zenodo](zenodo.org/)
+        - label: Under "Communities", tag with the-turing-way
+        - label: Share the Zenodo link and a recording (if available) in TTW Slack!
+ 


.github/ISSUE_TEMPLATE/give_a_talk.yml
```diff
@@ -0,0 +1,43 @@
+name: Give a TTW talk 
+description: Resources and checklist for giving a talk about the Turing Way
+title: "[TALK] <title>"
+labels: [talks-and-workshops, newsletter]
```
sgibson91, line 4:
```suggestion
labels: [talks-and-workshops, newsletter]
assignees: [aleesteele]
```

If we want to ping Anne every time this issue is opened so she can add details to the calendar, the easiest thing to do would be to assign her to it. How do you feel about that, @aleesteele ?

.github/ISSUE_TEMPLATE/give_a_talk.yml
```diff
@@ -0,0 +1,43 @@
...
+      placeholder: please use the format YYYY-MM-DD
+    validations:
+      required: true
+  - type: input
+    id: talk_details
+    attributes:
+      label: Details of the talk
+      description: Where are you presenting? Please provide a link to the event if available
+  - type: checkboxes
+    id: checklist
```
sgibson91, line 24:
```suggestion
    id: pre-talk-checklist
```

This form doesn't yet load correctly because each value of `id` must be unique

.github/ISSUE_TEMPLATE/give_a_talk.yml
```diff
@@ -0,0 +1,43 @@
...
+      label: Pre-talk checklist
+      description: Steps to prepare your presentation
+      options:
+        - label: Schedule your talk and let the community know in TTW Slack. If you’d like feedback and/or to schedule a practice talk, ask in the TTW Slack!
+        - label: Download template(s) from the [promotion pack](https://drive.google.com/drive/folders/1mzGmbJkPnP5q1goQesxDc_E5zAPL0eTF?usp=sharing) to ensure stylistic concistency
+        - label: If you re-used slides from a specific talk, please acknowledge the original author of the slides 
+        - label: Double check acknowledgements slide for TTW team, license info, Scriberia link, any additional acknowledgements
+        - label: Double check contact info for TTW links (book, Twitter, GitHub, Slack, newsletter), your own contact into
+  - type: checkboxes
+    id: checklist
```
sgibson91, line 35:
```suggestion
    id: post-talk-checklist
```

This form doesn't yet load correctly because each value of `id` must be unique

sgibson91 (commented):
There's a few bugs in rendering this form atm, a few of which I have provided fixes for.

I have also suggested how we can notify Anne that this issue template has been opened.

.github/ISSUE_TEMPLATE/give_a_talk.yml
```diff
@@ -0,0 +1,43 @@
...
+    attributes:
+      label: Post-talk checklist
+      description: Steps to share your presentation
+      options:
+        - label: Upload your slides as a PDF to [Zenodo](zenodo.org/)
+        - label: Under "Communities", tag with the-turing-way
```
sgibson91, line 41: 1. I think these two tasks should be merged, because selecting the community is part of the upload process
2. You could do this as a pre-talk task. In terms of accessibility, distributing your slides before the talk so people can download it to use with screenreading software, or just because, is really valuable

sgibson91:
Given that there's a pre- and post-talk checklist, I no longer feel confident on automatically closing these issues based on whether the date the talk was given has passed or not.

However, given that I've just advocated that 2 of the 3 tasks in the post-talk checklist be moved to the pre-talk checklist, maybe we don't need the idea of pre/post so much and we can go with our original plan?

.github/ISSUE_TEMPLATE/give_a_talk.yml
sgibson91: Also, maybe not use as a PDF. I usually do PDF and the original file format. It's a lot more difficult for others to reuse your slides if they're _only_ available in PDF format

sgibson91:
I am drafting the workflow in the below PR:

* https://github.com/alan-turing-institute/the-turing-way/pull/2571

.github/ISSUE_TEMPLATE/give_a_talk.yml
```diff
@@ -0,0 +1,43 @@
...
+        This template will guide you through the steps necessary to give a successful talk about The Turing Way!
+  - type: input
+    id: date
+    attributes:
+      label: Date of talk
+      description: When will you be presenting about the Turing Way?
+      placeholder: please use the format YYYY-MM-DD
+    validations:
+      required: true
+  - type: input
```
sgibson91, line 18:
```suggestion
  - type: textarea
```

I would make this a larger text area so users can provide more info if they would like. It's just a bit more freeform than the `input` type.

aldenc:
> 

agreed, although I do think it would be smart to not close it right away anyways, can we have it do something like close a week or 10 days later?

.github/ISSUE_TEMPLATE/give_a_talk.yml
aleesteele: I'm okay with that! I need to get my github notification management system together, as I currently get pinged for everything in the repo. This means that I accidentally lose things amidst all the notifications. If I stop those global notifications, and only get pings for things I'm tagged in, this should be okay @sgibson91?

sgibson91: This is really up to you, Anne! You will be receiving the "noise" so I would prefer to work this to fit _you_ rather than the other way around 💛

.github/ISSUE_TEMPLATE/give_a_talk.yml
```diff
@@ -0,0 +1,36 @@
...
+    attributes:
+      label: Checklist
+      description: Access TTW slides, check acknowledgements, and share your presentation
+      options:
+        - label: Schedule your talk and let the community know in TTW Slack. If you’d like feedback and/or to schedule a practice talk, ask in the TTW Slack!
+        - label: Download template(s) from the [promotion pack](https://drive.google.com/drive/folders/1mzGmbJkPnP5q1goQesxDc_E5zAPL0eTF?usp=sharing) to ensure stylistic concistency
+        - label: If you re-used slides from a specific talk, please acknowledge the original author of the slides 
+        - label: Double check acknowledgements slide for TTW team, license info, Scriberia link, any additional acknowledgements
+        - label: Double check contact info for TTW links (book, Twitter, GitHub, Slack, newsletter), your own contact into
+        - label: Upload your slides to [Zenodo](zenodo.org/), tag with the-turing-way under "Communities". Zenodo allows [versioning](https://help.zenodo.org/#versioning), so we encourage you to upload your slides before your talk and update the file with any changes after.
```
sgibson91, line 34:
```suggestion
        - label: Upload your slides to [Zenodo](zenodo.org/), preferably in the original format along with a PDF or any other format you are using.During the upload, tag with the-turing-way under "Communities". Zenodo allows [versioning](https://help.zenodo.org/#versioning), so we encourage you to upload your slides before your talk and update the file with any changes after.
```

Just making this point a little more specifc. WDYT?

aldenc: we were working on this at the same time! it's been updated

.github/ISSUE_TEMPLATE/give_a_talk.yml
aleesteele: @sgibson91 ❤️ A quick follow-up: I had a chat with @aldenc about this yesterday, and it might be a little tricky on the notification side of things. I currently get all alerts for the repository (that I can't turn off) because I'm an admin, which means that me specifically getting tagged in something might get lost in the wider list of notifications from the project. I think you are as well?

Thinking through alternatives: I wonder if there's a way for me to get pinged in a slack channel instead where github is integrated (i.e. everything tagged with 'newsletter' or 'talks-and-workshops' pings an alert into this channel? What do you think? I've been using the 'auto-github-updates' channel as a way of tracking the notifications in real time... I wonder if this could work?

sgibson91: It's definitely not true that you have to receive all notifications because you are admin of a repo. I believe it is the default, but you can always change these settings via the "Watch" button at the top of the page. So I **don't** get all notifications even though I am an admin. The "Custom" setting might be interesting to look into...

Screenshot 2022-08-16 at 16 44 02

That aside, that doesn't mean setting up a Slack channel is a bad idea if that is what you would prefer to do. If you want to create a specific channel for these notifications or continue to use the "auto-github-updates" one I'm happy to configure notifications there

sgibson91: Hmmm, the custom settings weren't as exciting as I'd hoped... (I was hoping for filtering by labels or created vs updated or something)

Screenshot 2022-08-16 at 16 51 56

Let's go with the Slack channel

sgibson91 (approved):
Yaaaay, let's goooo!

aleesteele (approved):

aleesteele:
🚀🚀🚀

Issue #2343, opened 2022-05-06
[BUG] 'master' in URL means that tests fail


da5nsy:
### Description
When text is added that includes a link to a URL that has 'master' in the URL, the tests fail (because of #2237, and seen currently in #2339). This is likely to be a common issue since until recently (still?) github used 'master' as the default name for the default branch in repos.

#### Proposal
I propose we update the behaviour of the bot such that 'master' is allowed if preceded by a '/'.

da5nsy:
Looks like there is no way to edit the regex at our end. I'll open an upstream issue because I think that this will be fairly widespread.

da5nsy:
Closed by #2588

PR #2588, opened 2022-08-18
[WIP] update in solidarity to allow for hyperlinks to repos that use "/master"

da5nsy:
Tweaking the 'in solidarity' settings to create a special case that allows the word "master" to be used, only in the situation whereby it is immediately preceded by a "/", as seen in https://github.com/alan-turing-institute/the-turing-way/pull/2339 and described in #2343.

Currently seeking two specific types of input:
- Is this a good idea?
- How do we test that this works? (I can't see how to test this without approving it and then purposefully doing other PRs which should pass/fail and seeing whether it works as intended, but I'd prefer to test it without having to approve this PR)

### DIFF
--- a/.github/in-solidarity.yml
+++ b/.github/in-solidarity.yml
@@ -1,6 +1,10 @@
 rules:
   master:
-    level: failure
+    level: off
+  master_excluding_when_with_preceding_slash:
+  regex:
+    - /(?<!\/)master/g
+  level: failure
   slave:
     level: failure
   whitelist:
--- a/.github/in-solidarity.yml
+++ b/.github/in-solidarity.yml
@@ -1,10 +1,6 @@
 rules:
   master:
-    level: off
-  master_excluding_when_with_preceding_slash:
-  regex:
-    - /(?<!\/)master/g
-  level: failure
+    level: warning
   slave:
     level: failure
   whitelist:


da5nsy:
Request @aldenc for review since they were the previous person to work on this

aldenc:
thanks for flagging @da5nsy 

I'm pretty sure we had to merge the original PR and then create a test case. tagging @sgibson91 in case she remembers differently, and also to please review the code changes

I think the alternative would be to change "master" from a failure to a warning so that we can allow it through in the case of a hyperlink

sgibson91:
> * How do we test that this works? (I can't see how to test this without approving it and then purposefully doing other PRs which should pass/fail and seeing whether it works as intended, but I'd prefer to test it without having to approve this PR)

Unfortunately with workflows like this, there's no way to test them until they're merged into the default branch. I usually test these sorts of changes by merging the change into the default branch of my fork first, and then making PRs in my fork to check the behaviour is as expected. (I also have a whole repo _just_ to test actions workflows 😅 )

sgibson91 (approved):
The code looks good to me

da5nsy:
As proud as I was of my custom regex I think actually the most parsimonious solution is just switching the master case from `failure` to `warning`.

PR #2636, opened 2022-09-13
added name to project members in ways-of-working.md

eirini-zormpa:
### Summary

This adds me as a project member in the ways-of-working.md

It's one line, so it should be straight-forward to review, but do look out for typos I guess!


EKaroune (approved):
thanks for updating!

PR #2728, opened 2022-10-14
Auto-format YAML with prettier and pre-commit

sgibson91:
### Summary

In https://github.com/alan-turing-institute/the-turing-way/pull/2726, there was an issue with the book build due to incorrectly formatted YAML. @aleesteele and I suggested it would be a good idea to enforce YAML formatting since it is a tricky thing to hand-edit and book build failures are not super informative.

* Configured [pre-commit](https://pre-commit.com/)
  * This is a tool that can be used locally to perform certain, specified actions just before git creates a commit
  * It can also be used in CI so that, for any open Pull Requests, if the files would have been edited when running pre-commit locally, an extra commit applying these changes would be added to the PR
* Configured [prettier](https://prettier.io/), which will be executed by pre-commit and perform the auto-formatting
  * Provided a `.prettierignore` file to restrict which files will be auto-formatted to only YAML. We can change this file as needed to include other file we would like to be auto-formatted.

This combination of pre-commit and prettier is a very common workflow among open source projects.

There will be three commits on this repo:

* The two original commits I made containing the pre-commit and prettier configuration files described above
  * [prettier config commit](https://github.com/alan-turing-institute/the-turing-way/commit/a2056934c4263ea5eed693c115d0ff691eab4900)
  * [pre-commit config commit](https://github.com/alan-turing-institute/the-turing-way/commit/39e3e9dc5b5ca3570aeaa7d31649d142f68c4ba4)
* An automatically generated commit applying prettier auto-formatting to YAML files, by pre-commit
  * [link TBA]

- [ ] Does everything look ok?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,21 @@
+# pre-commit is a tool to perform a predefined set of tasks manually and/or
+# automatically before git commits are made. pre-commit can also be run in CI
+# on Pull Requests and will auto-commit corrections to the PR for consistency.
+#
+# Config reference: https://pre-commit.com/#pre-commit-configyaml---top-level
+#
+# How to install locally
+#
+# - pip install pre-commit
+#
+# Common tasks
+#
+# - Run on all files:   pre-commit run --all-files
+# - Register git hooks: pre-commit install --install-hooks
+#
+repos:
+  # Autoformat: yaml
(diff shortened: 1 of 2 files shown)


sgibson91:
Pre-commit doesn't auto-add correction commits for prettier and it seems this is intentional (see refs below). So this approach will only work for those using pre-commit locally.

PR #2745, opened 2022-10-21
Create book_dask_template.yml

aldenc:
### Summary

yaml template for bookdash checklist for organizers

### DIFF
--- a/.github/ISSUE_TEMPLATE/book_dask_template.yml
+++ b/.github/ISSUE_TEMPLATE/book_dask_template.yml
@@ -1,0 +1,156 @@
+name: "\u1F4DA Book Dash Checklist"
+description: Book Dash Planning Checklist
+title: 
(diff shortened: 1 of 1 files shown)


aldenc:
@aleesteele made this yaml version of the book dash template create here: https://github.com/alan-turing-institute/the-turing-way/pull/2734

I reorganized it into smaller sections since yaml can't handle indented sub-lists in the checkboxes, and tried to clarify some of the points. I think it will be really good if we can add a timeline throughout, eg when is the 1st planning meeting, when do applications get reviewed, etc. I've flagged the sections where we could add a timeline.

aldenc:
added fireside chat issue template

aleesteele:
Tagging @AlexandraAAJ here for review - as we plan for the May 2023 Book Dash, we can review this template together!

Issue #1313, opened 2020-08-26
[accessibility] Reduce image sizes to lower than 1 mb  ( 26 PR ideas for HacktoberFest participants)


malvikasharan:
## :tada::balloon::cake: **Welcome Hacktober Contributors! This is The Turing Way repository** :cake::balloon::tada:

:dizzy::hatched_chick::sunny: _We're excited that you're here and want to contribute._ :sunny::hatched_chick::dizzy:

### About Hacktober and The Turing Way
Hacktober is a month long celebration of Open Source Projects where the goal is to make 4 Pull Requests during the month.

#### A bit about Git
If you’re new to git then here are [some instructions](https://help.github.com/articles/signing-up-for-a-new-github-account/) to help you get going. Remember that you can ask us any questions you need to along the way.

A [Pull Request](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request) (what Hacktober is measured on) lets you tell others about changes you've pushed to a branch in a repository on GitHub. Once a pull request is opened, you can discuss and review the potential changes with collaborators and add follow-up commits before your changes are merged into the base branch.

Under the Issues listed in our Hacktober tag, we’ve outlined what needs to be done, and how you can go about addressing it. Once the Issue has been addresses, you can submit a Pull Request for review (tell us what you’ve done) and we can incorporate it into The Turing Way.

We’ve also created this special Hacktoberfest tag to make is easy for you to find places to contribute if you’re new.

In the spirit of Hacktoberfest, please only do a few of these to submit your PR so that we can learn and grow as a community together.

#### Our Community
Please read our core Contributors document and our Code of Conduct while you engage in The Turing Way. We want to ensure that every user and contributor feels welcome, included and supported to participate in The Turing Way community.

In addition, for Hacktober we would also like to recognise the work that you are doing. If you would like to be featured on our [Twitter](https://twitter.com/turingway?s=20) Hacktoberfest series then: 
- Add your twitter handle to the comments when you submit your PR, 
- What brings you to Open Source, Hacktoberfest, and/or The Turing Way,
- and anything else you’d like to celebrate about your contribution!

We welcome all contributions to this project via GitHub issues and pull requests. Please follow these guidelines to make sure your contributions can be easily integrated into the projects. As you start contributing to The Turing Way, don't forget that your ideas are more important than perfect pull requests. ❤️

[...]

PR #2767, opened 2022-11-05
Add SVG images

da5nsy:
### Summary

Reducing the filesize of images rendered in the book means that the book can be browsed over poorer internet connections.
In the past, this has been achieved through lossy image compression (See #1313) but this reduces the visual quality of images.
For a subset of [the Scriberea images](https://doi.org/10.5281/zenodo.3332807) copies in SVG format exist (`zz-TheTuringWay-Scriberia-2019-20-SVG-where-available.zip`). 
This is a vector graphic format, which means that the images can be rendered at any size, with no loss of quality. 
For illustrations with lots of block colors, this format is also very storage efficient, often outperforming a lossily compressed jpeg.

- Add the SVG images to the repo.
- Modify the references to all images such that those with SVG versions use the SVG version.
   - I do this by replacing all `filename.jpg` and `filename.png` with `filename.*`, in which case Jupyter Book chooses which to use, and it prioritizes SVGs.
   - I chose to do this rather than hard code the file extensions, so that it will be easier in the future to change the format of an image (one will not have to change the references to that image).
   - It also means that if jupyter book is used to generate a different type of output that doesn't allow SVGs, then it will fall back on the other filetype.
- Update the style guide to mention the `filename.*` usage.
- 37eb85d compresses the largest remaining bitmap images
   - in most cases, images were converted to indexed-color pngs (with ~10 locally indexed colors)
   - where gradients had been used, pixel dimensions were reduced (to long edge = 2000px) and jpeg compression was used

- [ ] Check to see whether any images are broken
- [ ] Try different browsers and hardware to check that the SVGs render well
- [ ] Are there more vector graphics that could be added? For more recent images there are jpgs and pdfs (and the pdfs seem to be bitmaps, disappointingly)


da5nsy:
37eb85d closes #1313

da5nsy:
_Reminder to **squash and merge** this PR, to avoid adding and then removing `legal-restrictions.svg`_

da5nsy:
It would make me happy to also change over the images also used in `.github\config.yml` to be the SVG versions, since these would look ever so slightly nicer, but they are hardlinked to the zenodo versions, and the SVGs on zenodo are zipped so... 🤷

EstherPlomp (approved):
I think this looks good! Very handy and some of the pages indeed load faster now <3 Thanks Danny for going over everything - must have been a lot of work!

da5nsy:
Thank you @EstherPlomp for reviewing! ❤️

PR #2773, opened 2022-11-07
[CI] Handle cases where talk_date cannot be extracted from the issue body

sgibson91:
### Summary

In the Close Talk Issues workflow, handle cases where `talk_date` cannot be extracted from the issue body, causing errors such as:

https://github.com/alan-turing-institute/the-turing-way/actions/runs/3406596179/jobs/5665400691#step:4:136

- [ ] Everything looks ok?

### DIFF
--- a/.github/workflows/close-talk-issues.yml
+++ b/.github/workflows/close-talk-issues.yml
@@ -83,20 +83,24 @@ jobs:
                   # Extract date from issue body
                   match = pattern.search(issue.body)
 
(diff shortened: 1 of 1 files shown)


sgibson91:
> Maybe it would be better to have workflows like this execute a script (in `.github/workflows/scripts/` or similar) rather than writing the python inline.

Yeah, I go either way on this depending on the workflow ⚖️

PR #2861, opened 2022-11-17
Add continuous integration so that build fails on jupyter-book warnings

da5nsy:
### Summary

Implements #2838 

It might also be worth considering the `--keep-going` flag. I'm not including it here because:
- hopefully once this is implemented there will only be one error at a time
- it means that it will fail quicker, so people get feedback quicker, and also less compute time
- If this _doesn't work_ there's less things for us to debug right now (and we can consider adding it in later)

docs reference: https://jupyterbook.org/en/stable/basics/build.html#debug-your-books-build-process

- [ ] If there are strong feelings we could add `--keep-going` flag already
- [ ] Does it mean that CI fails on new commits (it should, because there are a small number of outstanding warnings - see #2611 and #2838)

I believe everyone is already in contributors, but for the record: thanks to @sgibson91, @bsipocz, @choldgraf and others for input  

- [ ] All contributors to this pull request are already named in the [table of contributors](https://github.com/alan-turing-institute/the-turing-way/blob/main/README.md#contributors) in the README file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Build Jupyter-Book
         run: |
           cd ./book/website
-          jupyter-book build . -W
+          jupyter-book build . -W --keep-going
 
       - name: Run html proofer
         if: matrix.os == 'ubuntu-latest'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Build Jupyter-Book
         run: |
           cd ./book/website
-          jupyter-book build .
+          jupyter-book build . -W
 
       - name: Run html proofer
         if: matrix.os == 'ubuntu-latest'


da5nsy:
I've never been happier to see CI fail 🤣

sgibson91:
I would dive in for the keep going flag now so we at least know if it work the way we expect it to. We can always drop the commit of it doesn't :)

da5nsy:
> I would dive in for the keep going flag now so we at least know if it works the way we expect it to. We can always drop the commit of it doesn't :)

Related to this discussion, do you think `fail-fast: false` is still necessary? https://github.com/alan-turing-institute/the-turing-way/commit/a61731182d3fcc1a71f21049d315f5865b47282f
I don't know to what extent limited compute time is for TW (💰 ?) but I don't personally see much benefit in keeping it

da5nsy:
> I would dive in for the keep going flag now so we at least know if it works the way we expect it to. We can always drop the commit of it doesn't :)

It works! 🥳 

https://github.com/alan-turing-institute/the-turing-way/actions/runs/3491516828/jobs/5844240457
image

bsipocz (approved):

bsipocz:
> fail-fast: false

Yes, I think it's good to keep it here given none of the jobs is taking really long to run, and they are very distinct. (I use `fail-fast: true` for cases where we know that a job takes a long time, so we only trigger it when all else is fine, or when we know that is one job fails another is likely fail, too, etc.)

Issue #2897, opened 2022-11-28
[BUG] A colon symbol appears in the toc when it is viewed


RayStick:
# Description/Actual behaviour

_I think this is a 🐛 ?_

Just recently merged this PR - https://github.com/alan-turing-institute/the-turing-way/pull/2883
The [toc file](https://github.com/alan-turing-institute/the-turing-way/blob/main/book/website/_toc.yml#L538) says:

              - title: Data Wrangler Overview
                file: collaboration/research-infrastructure-roles/data-wrangler

Strangely, the way it shows up in the toc is like this (below) which seems to match h1 in the file itself, not what is in the _toc.yml
See the colon symbol in "Data Wranglers: Overview" which does not match the others

<img width="1008" alt="Screenshot 2022-11-28 at 12 13 32" src="https://user-images.githubusercontent.com/50215726/204276068-8448ace3-444a-45c3-b140-554b4703296b.png">

# Expected behaviour

Expected it to look like this. Even though h1 in Data Stewards has a colon, the toc display down the side does not. 

<img width="1047" alt="Screenshot 2022-11-28 at 12 13 43" src="https://user-images.githubusercontent.com/50215726/204278684-365bfd92-1805-4815-a5c6-2d3397291482.png">

RayStick:
This is what we discussed in the co-working call @EKaroune, @aleesteele, @EstherPlomp and @harisood

da5nsy:
This is a weird one. Thanks for flagging @RayStick.

I can reproduce this locally.
If I modify other TOC entries, they update, but if I modify this one it does not (even when I force a full rebuild with `jupyter-book build --all website`)

I'm stumped.

sgibson91:
Hi @RayStick - This is not a bug, the title has been defined differently in two places: the `_toc.yml` file and the Markdown file. There is a colon in the title of the Markdown file created here: https://github.com/alan-turing-institute/the-turing-way/pull/2883/files#diff-5457e1a7583532adcc2da57ff2e270232267393fcef476fe2c9a5a17beef922cR2

As you pointed out with your screenshot, there is no colon in the title on the right hand side menu because that is pulling the data from the `_toc.yml` file. The main page, however, is pulling the content of the Markdown file directly.

There are two ways to achieve consistency here, either of which are acceptable:

1. Remove the colon from the title defined in the `book/website/collaboration/research-infrastructure-roles/data-wrangler.md` file, resulting in no colons anywhere, or
[...]

PR #2903, opened 2022-12-01
[do not merge] more debugging of #2897

sgibson91:
Remove a colon from an alt title entry

Expected behaviour is that the colon will be removed in the sidebar

related #2897


sgibson91:
Confirming that we see the expected behaviour in the preview

Screenshot 2022-12-02 at 11 24 54

PR #2905, opened 2022-12-02
[do not merge] more testing of TOCs

sgibson91:
Extension of #2903 

See if the removal of double quotes replicates the behaviour reported in #2897


sgibson91:
Nope, this one also does not replicate the behaviour of #2897 either

Screenshot 2022-12-02 at 15 56 23

Issue #1172, opened 2020-05-25
Correct all broken links listed in this issue


malvikasharan:
### Summary

* Upgrading Jupyterbook led to many broken links, which needs fixing

### What needs to be done?

- [ ] Correct broken links that will be soon listed below
- Please mark the links that are fixed

### Who can help?

* This issue is being created for the new contributors, please correct maximum 3-5 broken link so that others can also contribute to fixing these

### For the maintainers

- [ ] We have disabling the html proofer until all links are corrected (see #1171) - once all the links are fixed, please activate them again on the CI workflow
---

### Updates

* I will add this on 26 May 2020

EstherPlomp:
@da5nsy I think this issue is outdated with the new workflow to check links? Can we close this?

da5nsy:
Hmm, yes and no. 
It is outdated and should be closed 👍 
Looking into this has alerted me to the fact that the link checker does not seem to be working as intended though 👎  
(I'll make a new issue regarding that though)

PR #2947, opened 2023-01-24
[WIP] Testing link checker CI workflow

da5nsy:
### Summary

Following the closure of #1172, I checked the link-checker workflow and saw that there were lots of errors, but that we were not being alerted to the errors.

See [here](https://github.com/alan-turing-institute/the-turing-way/actions/runs/3856037595/jobs/6571763097) for example.
There are broken internal links, broken external links, and timeout errors...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,5 +57,5 @@ jobs:
         uses: chabad360/htmlproofer@master
         with:
           directory: "./book/website/_build/html"
-          arguments: --assume-extension true
+          arguments: --assume-extension --disable-external --only_4xx
         continue-on-error: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,5 +57,5 @@ jobs:
         uses: chabad360/htmlproofer@master
         with:
           directory: "./book/website/_build/html"
-          arguments: --assume-extension
+          arguments: --assume-extension true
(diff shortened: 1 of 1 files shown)


sgibson91:
Ummm, I'm not sure how to review this PR as there's nothing in the "Files changed" tab?

Screenshot 2023-01-25 at 10 53 55

sgibson91:
I also remember a while ago that the link check action we are using here caused me some trouble on my website repo. I think I ended up switching to a different tool entirely.

JimMadge:
Some thoughts...

I have used/do use html-proofer on a few projects to check links in html ([here](https://github.com/alan-turing-institute/REG-handbook/blob/bc7b508ae55a564eff39fe0c6eca71e68b116a10/.github/workflows/ci.yaml) and [here](https://github.com/alan-turing-institute/data-safe-haven/blob/aef639f5fd5204c66f0f6e4cb89c844100791ff5/.github/workflows/lint_code.yaml)).
Major version updates (and breaking changes) happen.
Some URLs fail often or always even when they are OK (might be authentication issues, might be some sort of DDOS protection).
You need to be a bit diligent to keep a list of excluded domains up to date.
Random failures will also happen so anyone reviewing a PR needs to be aware and check for false negatives.

I've not seen lychee before. Looks interesting. Might help simplify CI where I have a checker for Markdown and one for HTML.

PR #3030, opened 2023-04-06
Add references.bib to the ignore file for the no-bad-latin checker

da5nsy:


adswa (approved):
LGTM - thanks much! To add context to this PR for anyone else reading it: In the context of #3028, a bibtex entry contained the word "etc", which the latin linter stumbled upon. Since the contents of the bibtex entries are neither part of the rendered chapters contents, nor in all cases changeable (think for example a latin phrase in the title of a paper), we felt that it made sense to exclude this file from being linted.

PR #3034, opened 2023-04-06
case sensitive rename

da5nsy:
`Adding-glossary.gif` starting with a capital somehow snuck through CI in #2961, but is [now causing CI to fail](https://github.com/alan-turing-institute/the-turing-way/actions/runs/4631361443/jobs/8194127498?pr=3028).

I had also intended to delete Translation.jpg (partly because it is capitalised, and partly because there is an equivalent svg version) in #2961 but it seems like that got missed or reverted.


adswa (approved):
I can confirm that this fixes the issue locally for me, let's wait for CI to agree on that! Thanks for the quick fix!

PR #3041, opened 2023-04-11
image formatting (minor tweaks)

da5nsy:
Following up to #2950 with some edits that I couldn't include because they clashed with other PRs.

- Reduced size of some images
- Removed specific file format requests in some text
- Removed left over drafts from some pages so that future searches for file format requests are easier


bsipocz (approved):
All looks good. Thanks!

PR #3093, opened 2023-05-22
Automate updating contributor records in the book

JimMadge:
### Summary

Contributes to #2988 

These changes should mean that the books record of contributors is updated as soon as the relevant 'source' files are updated, rather than on a weekly basis.

- Update `all-contributors.rc` so that the bot (and CLI) write the contributors table to `README.md` and `book/website/afterword/contributors-record.md`
- Use MyST `include` directive to insert `contributors.md` into `book/website/afterword/contributors-record.md`
- Remove script and workflow to do the above on a weekly basis
- Adjust header levels so that contributor records doesn't have to `<h1>` headings
- Adjust cross references
- Add documentation for how contributors are rendered in the book

- [ ] Everything looks ok?

- [ ] All contributors to this pull request are already named in the [table of contributors](https://github.com/alan-turing-institute/the-turing-way/blob/main/README.md#contributors) in the README file.

### Updates

- Documentation changes that were originally in #3102 are now here.


.all-contributorsrc
```diff
@@ -1,6 +1,7 @@
 {
   "files": [
-    "README.md"
+    "README.md",
```
sgibson91, line 3: Can I recommend that we add a link to the configuration docs as a comment to the top of this file so that folks working on this stuff know where to go for info? As I had to look up the `files` key to understand that it's purpose and your intention matched 😄 

Relevant link is: https://allcontributors.org/docs/en/bot/configuration

sgibson91 (approved):
One small recommendation, otherwise LGTM

.all-contributorsrc
JimMadge: Good point, I should do the same in #3102.

Turns out JSON doesn't have comments in the same way as programming languages. I added a key called `_comment` with the information and the CLI doesn't complain and still parses the file.

Fixed in f8b673b.

JimMadge:
@malvikasharan I want to run this past you before merging because it will change the way contributors are added to the book (but for the better!).

.all-contributorsrc
sgibson91: Did `//` not work? What failed exactly?

JimMadge: With the CLI `//` does work, so I would guess that works with the bot too. I think it is a bit risky though as it is not valid JSON so a parser could quite reasonably return an error.

That said, there is nothing special about an underscore to stop a program reading and using `_comment` :shrug:.

JimMadge:
@aleesteele I have made changes like you suggested in #3102 here as that part of the documentation was moved.

PR #3192, opened 2023-06-07
Create Newsletter_41_Mar2023.md

AlexandraAAJ:
Adding the March Newsletter.


malvikasharan (approved):
Looks good. Thanks Alex. We should add the image files too.

PR #3202, opened 2023-06-15
Add link to Chinese readme

da5nsy:
The link to the Chinese readme appears to have been accidentally excluded until now.


EstherPlomp (approved):
Oops! Thanks for spotting and fixing @da5nsy!

PR #3206, opened 2023-06-21
Suggest a more recent version of numpy to install in z2b tutorial

sgibson91:
numpy v1.14.* no longer works on mybinder.org when using the Python version set by default. This is because the Python version has been upgraded beyond what will support this old version of numpy. So, we recommend a newer version of numpy to install instead which is compatible with the Python default in mybinder.org.


aleesteele (approved):
Thank you Sarah! 🚀

PR #3209, opened 2023-06-28
Slight update of pd-overview-sharing.md

malvikasharan:
@jcolomb I should have suggested these changes before approval - can you check and see if this makes sense.


jcolomb (approved):
yes make sense

PR #3288, opened 2023-09-13
[DO NOT MERGE] test - do not merge

da5nsy-test:
Just testing the new user workflow


da5nsy-test:
Here is an image without alt-text.

da5nsy-test:
Here's another image without alt-text. This time a jpg rather than an svg.

bsipocz:
This test PR is a 1+ year old, so I go ahead and close it. Danny can always open a new one for any new tasks that needs to be tested this way.

PR #3392, opened 2023-11-13
Rationalise cross referencing examples

### LINKED ISSUE #3349: [TO EDIT] community-handbook/style/style-crossref streamline example cases
### Summary

On the "community-handbook/style/style-crossref" page, ["Examples of cross-referencing"](https://the-turing-way.netlify.app/community-handbook/style/style-crossref#examples-of-cross-referencing) section, there are several cases provided for how to include cross references.

The cases are really clear and helpful; however looking carefully at the following three cases:

1. Cross-referencing a section within the same file before the reference.
2. Cross-referencing a section within the same file after the reference.
3. Cross-referencing a section within a different file.

It looks like (unless I'm missing something, which I may well be) the three cases use the same approach. 

I was wondering  if the advice might be simpler and made clearer if the three cases were combined into a single case, so that the question of whether a reference appears before, after or on a separate page can be avoided.

This is just an idea that occurred to me while reading through the section, which I found to be generally very clear and helpful.

### What needs to be done?

 - [ ] Consider whether there's a reason for having three cases that I missed.
 - [ ] Consider whether the change makes things clearer, or less clear.
 - [ ] Combine the three cases into a single case, noting that "it doesn't matter whether the label appears before or after the reference, or even on a completely different page" (or similar).

### Who can help?

From the git log it looks like the examples cases were originally added by @pgadige in commit 438de8a02bf2296b21cff1f96e5ae0d46bacc0b5. It would be especially helpful therefore to get @pgadige's opinion on this proposal.

### Updates

- [ ] First set of edits to chapter/subchapter
- [ ] Proofread
- [ ] Request reviews
- [ ] Address reviews
- [ ] Merge to main branch

llewelld:
### Summary

Rationalises the examples of cross referencing in the `community-handbook/style/style-crossref section`.

Since the relative locations of the reference and citation don't affect the syntax, the three examples have been combined into a single example to help simplify the text.

Fixes #3349.

I also considered @kallewesterling's nice suggestion to add an example that includes a label, but noticed that it looks like this case is already covered in the section on the page ["Providing an alternative title for the references"](https://the-turing-way.netlify.app/community-handbook/style/style-crossref#providing-an-alternative-title-for-the-references), so I think there's no need to add this example. However another possibility might be to combine the section below into the examples.

* Combines the four examples provided into two, by avoiding the issue of where the label is relative to the citation.

- [ ] Does the change make creating cross-references less clear?
- [ ] Does the updated text read okay?
- [ ] Does the page still render correctly?

### DIFF
--- a/book/website/community-handbook/style/style-crossref.md
+++ b/book/website/community-handbook/style/style-crossref.md
@@ -80,54 +80,13 @@ This label can be created in the corresponding file for the suggested section na
 
 ### Examples of cross-referencing
 
-**Examples for cross-referencing sections of chapters and subchapters**
-
+Here we provide examples of how to cross-reference chapters, sections of chapters and subchapters.
 We will use examples for the chapters in the "Reproducible Research" guide located in the `book/website` directory.
 
-**_Case 1_**: When you cross-reference a section of the chapter within the same file _before_ a label has been created.
+**_Example 1_**: Cross-referencing a chapter or subchapter (chapter/subchapter).
 
-Taking the previous example of `rr-overview-resources-addmaterial`, we can use this label to cross-reference
-it in an earlier section within the same file using the following:
-
-```
-{ref}`rr-overview-resources-addmaterial`
-```
-
-This will appear in the online book like so: {ref}`rr-overview-resources-addmaterial`.
-
-**_Case 2_**: When you cross-reference a section of the chapter within the same file _after_ a label has been created.
-
-In the same subchapter "Resources", we have created a label `rr-overview-resources-reading` for the section "Further Reading".
-We can cross-reference it in a later section within the same file using the following:
-
-```
-{ref}`rr-overview-resources-reading`
-```
-
-It will appear in your chapter like this: {ref}`rr-overview-resources-reading`.
-
-**_Case 3_**: When you cross-reference a section of a chapter in a different file (chapter) before or after a label has been created.
-
-In the subchapter "Definitions" of the "Overview" chapter, we have created a label
-`rr-overview-definitions` for the section "Table of definitions for reproducibility".
-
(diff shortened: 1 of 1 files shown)


kallewesterling (approved):
I think this looks good.

Arielle-Bennett:
oh dear, looks like merging the rest of the book broke this PR :weary: particularly frustrating without any specific examples....

llewelld:
I think the two checks might have failed because there were no runners available.

> The hosted runner: GitHub Actions 2 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

Is it possible to retrigger them? I don't see a way to do it (maybe I don't have the permissions?).

kallewesterling:
> oh dear, looks like merging the rest of the book broke this PR 😩 particularly frustrating without any specific examples....

Yes, there's no reason why we should merge the main into each PR before merging it back with the main. I'd probably try to do a git reset to the commit before (https://github.com/the-turing-way/the-turing-way/pull/3392/commits/73665e3c52f0f6278f907e910efc8b73e46a334e)

llewelld:
Ah, I can reset, but in that case I don't have the ability to merge (the only options I have are "Update with merge commit" and "Update with rebase").

llewelld:
Reset to remove the merge commit.

llewelld:
If someone could merge this for me I'd be grateful (I don't have the write permissions to do it myself).

Arielle-Bennett:
> If someone could merge this for me I'd be grateful (I don't have the write permissions to do it myself).

Fixed that, you should be able to now 😄

llewelld:
Thank you @Arielle-Bennett, it looks like that did fix it!

Issue #3357, opened 2023-11-13
[TO EDIT] communication/social-media update Twitter references


llewelld:
### Summary

In various places throughout _The Turing Way_ there are references to Twitter. These come in multiple forms:
1. Links or references to tweets.
2. Use of Twitter in relation to promotion, for example of events run by The Turing Way.
3. Discussions about how to use social media platforms effectively.

The largest sections related to Twitter are probably those in the [Social Media for Research Communications](https://the-turing-way.netlify.app/communication/social-media.html) sections.

Since Twitter recently changed its name to X, many publications now refer to Twitter as "X (formally Twitter)" or something similar. Moreover some functionality has changed, including the ability for users without X accounts to view material on the platform.

It might make sense therefore to update references to Twitter to use the "X" name instead, and to update the content around Twitter to incorporate recent changes to the platform.

This relates to, but is separate from #3212. That issue is primarily concerned with links to Twitter, whereas this issue is about how to refer to Twitter and the material that relates to use of Twitter.

### What needs to be done?

1. [ ] Update Twitter references to use the "X" branding instead in a sensible way.
2. [ ] Review the material in the [Social Media for Research Communications](https://the-turing-way.netlify.app/communication/social-media.html) sections to ensure it still applies.
3. [ ] Update the material so that it better reflects the current reality.

### Who can help?

1. Much of the material in these sections appears (according to the git log) to have been authored by @malvikasharan, @EKaroune and @EstherPlomp.
2. Anyone else with an interest in Twitter and how it's used.

### Updates

- [ ] First set of edits to chapter/subchapter
- [ ] Proofread
- [ ] Request reviews
- [ ] Address reviews
- [ ] Merge to main branch

llewelld:
According to grep, all of the following files reference "twitter":
```
$ grep -rInil "twitter" * --include="*.md" --exclude-dir="_build"
afterword/contributors-record.md
collaboration/organising-conference.md
collaboration/organising-conference/reg-form-template.md
collaboration/leadership/leadership-story-data-librarian.md
collaboration/new-community/new-community-guide.md
collaboration/informal-chats.md
collaboration/event-participation.md
collaboration/team-manual.md
collaboration/research-infrastructure-roles/rse-personal-story.md
communication/social-media.md
communication/social-media/social-media-twitter-tips.md
communication/social-media/social-media-twitter-multiple.md
communication/aa/aa-stories-community.md
communication/aa/aa-stories-interdisciplinary.md
[...]

Issue #3359, opened 2023-11-13
[TO EDIT] communication/social-media add Mastodon


llewelld:
### Summary

Currently the [Social Media for Research and Communication](https://the-turing-way.netlify.app/communication/social-media) section includes material on the following platforms:
1. Twitter.
2. LinkedIn.
3. Facebook.
4. Instagram.

The subchapters have a particular focus on Twitter:
1. [Tips for Starting with Twitter](https://the-turing-way.netlify.app/communication/social-media/social-media-twitter-tips).
2. [Managing Multiple Twitter Accounts](https://the-turing-way.netlify.app/communication/social-media/social-media-twitter-multiple).

Given recent changes to the way Twitter works (for example, the restriction on viewing tweets unless you have an account, which in some regions is also now paywalled), many academics and researchers are switching to open, federated alternatives such as Mastodon.

The Turing Way also now has [solid presence](https://fosstodon.org/@turingway) on Mastodon which isn't always well reflected in the text of the book.

There are already some issues related to this:
1. #3212: Twitter links
2. #2811: Using/supporting open infrastructure - suggested alternatives to currently used tools 
3. #3357: communication/social-media update Twitter references

In addition to these it would also be great if some material about Mastodon could be added to the section on Social Media usage. This would complement rather than replace the existing material about Twitter and other social media platforms.

There is already a lot of great material out there on how to use Mastodon effectively, and also in academic/research settings, that could be used for inspiration:
1. Bellini Saibene, Yanina Noemí; Crall, Alycia; Masinyana, Oscar; ["Mastodon Quick Start Guide"](https://zenodo.org/records/10019853). Resource aimed at science, education and open source.
2. v buckenham, ["CW: Mastodon"](https://v21.io/blog/uses-of-cws). How to use Mastodon Content Warnings effectively.
3. ["Fedi.Tips"](https://fedi.tips/about-this-site/). An Unofficial Guide to Mastodon & the Fediverse. Covers a diverse range of topics; not science specific.

I appreciate these issues may have been raised and discussed already and some decisions made about them, in which case this can be closed. I looked for other related tickets but couldn't see anything directly related to this topic.

### What needs to be done?

1. [ ] Decide on whether it makes sense to incorporate material on Mastodon.
2. [ ] In the communication/social-media chapter add a small section to discuss Mastodon alongside the other four social media platforms mentioned.
3. [ ] Consider adding more subchapters related to use of Mastodon for data science.

### Who can help?

[...]

PR #3396, opened 2023-11-13
[Ready for review] Update Twitter references to use X

llewelld:
### Summary

Since Twitter changed its name to X, and Tweetdeck changed its name to X Pro, the text needs updating to reflect this. These changes try to balance between the two.

I'm not entirely convinced this is a good idea. Or at least, I think keeping up with the X rebranding is probably a bad idea but still may be necessary.

I've tried to retain the link back to Twitter ("X (formerly Twitter)") at least once on each page that I changed. I've also taken the opportunity to update some of the details, such as the fact that X Pro (formerly Tweetdeck) is now only available to paid subscribers.

Although this is a work-in-progress, I think it already gives a pretty clear idea about the nature of the changes, so I'd really welcome input on whether this should be done, or if there's a better way to approach it.

Contributes to #3357.

* Update Twitter references to talk about X instead.
* Update the info related to Twitter (such as image sizes, paid for services, tweet lengths and so on).

- [ ] Is this worth doing at all?
- [ ] The approach is to talk about X, but mention Twitter ("formerly Twitter") on the first mention on a page. Is this a good approach? Is there a better approach.
- [ ] Are the updates related to how twitter works correct?


kallewesterling (approved):
I like the changes you've made here @llewelld. But I agree, I wonder if we want to have some kind of statement somewhere about how the platform isn't what it once was for academics/researchers, and that many have left it, in favour of Mastodon or BlueSky as alternatives.

llewelld:
Thanks for the feedback @kallewesterling. I've now gone through and caught all of the Twitter references I could find.

I like your suggestion about "the decline of Twitter" and so have added some material about that too. I didn't mention Mastodon as I'd planned to cover that in #3359, but could certainly add something about that here in preparation if you think that would work better?

KirstieJane:
I think this is SUCH an interesting PR - I'm delighted that you've opened it @llewelld! 

When I first read it my gut instinct was "uuuuuurgh, no, why can't we just keep calling it Twitter". And I do have a decent part of me that feels that way.

BUT - it isn't a particularly justifiable position. (It's a very emotional one! I miss Twitter of 5-10 years ago!) And one of the goals of having an online book is to keep the chapters up date over time.

So I think I like the changes. I particularly appreciate keeping the "formally Twitter" text on each page.

Keen to hear if others have strong opinions.... thank you @llewelld and @kallewesterling for developing the PR!

llewelld:
Thanks for your very thoughtful feedback @KirstieJane. I think I'm in the same place as you when it comes to Twitter :smiley: and also eager to hear any other opinions on this.

kallewesterling (approved):
This definitely captures what I was saying in my previous review + with an academic reference to boot! Excellent work! 😄  👍

kallewesterling:
I should also add that this could be changed to `[ready for merge]` in the title, in case you're happy with all the changes, @llewelld but perhaps we should let some conversation happen throughout the remainder of the book dash? @KirstieJane ?

EstherPlomp:
Thanks so much for going over all of this @llewelld!! 

I'm also still grieving a bit about Twitter, but it is what it is and it is great that the content of the book will be updated to reflect that!

book/website/communication/social-media.md
```diff
@@ -20,40 +20,40 @@ Although, maintaining a social media account successfully can be time-consuming.
...
 It has become the communication channel of choice for many researchers and is opening up research discussions at a much earlier stage than previously seen.
 
-Twitter can be used to build up a personal academic profile, for use throughout a research project and can also be used by organisations to promote their work and message.
-It can make research more visible and understandable to wider communities by allowing non-scientists to find new research instantaneously.
+X can be used to build up a personal academic profile, for use throughout a research project and can also be used by organisations to promote their work and message.
 Scientists can also be contacted directly by the public, which enables a more approachable level of communication {cite:ps}`Kelesidou2021scicomm`.
+However recent changes to the platform have also alienated some academics and in the longer term this may precipitate a more widespread move by researchers away from X to more open platforms {cite:ps}`VidalValero2023twitter`.
```
Arielle-Bennett, line 37:
```suggestion
However recent changes to the platform have also alienated some academics and in the longer term this may result in a more widespread move by researchers away from X to more open platforms {cite:ps}`VidalValero2023twitter`.
```

Suggesting slightly clearer language to help with translations and non-fluent English speakers

book/website/communication/social-media.md
```diff
@@ -20,40 +20,40 @@ Although, maintaining a social media account successfully can be time-consuming.
...
 
-Due to its popularity and wide range of uses for academic purposes, the next two sub-chapters focus on using Twitter.
-However, below is an overview of the most popular general social media sites used by researchers - Twitter, Linkedin, Facebook and Instagram.
+Despite this X remains popular and given its wide ranging uses for academic purposes the next two sub-chapters focus on how to use it effectively.
```
Arielle-Bennett, line 39:
```suggestion
Despite this X remains popular and, given its wide ranging uses for academic purposes, the next two sub-chapters focus on how to use it effectively.
```

book/website/communication/social-media.md
```diff
@@ -20,40 +20,40 @@ Although, maintaining a social media account successfully can be time-consuming.
...
+Below we also offer a brief overview of the most popular general social media sites used by researchers - X, LinkedIn, Facebook and Instagram.
 
-### Twitter
+### X (formerly Twitter)
 
 * They have 353 million monthly active users.
```
Arielle-Bennett, line 44: I wonder if we should delete this? Number of active users has fallen dramatically and this figure is almost definitely out of date now

book/website/communication/social-media/social-media-twitter-tips.md
```diff
@@ -1,20 +1,20 @@
 (cm-social-media-twitter-tips)=
-# Tips for Starting with Twitter
+# Tips for Starting with X
 
-Twitter has 353 million monthly active users.
-You are limited to 280 characters and the style is more informal and chatty than other platforms like Linkedin.
-You can use links, photos (1200 x 628 pixels) and videos (max 2 minutes 20 secs) in similar ways to the other social media platforms.
+X (formerly Twitter) has 353 million monthly active users.
```
Arielle-Bennett, line 4: I wonder if we should delete this? Number of active users has fallen dramatically and this figure is almost definitely out of date now

book/website/communication/social-media/social-media-twitter-tips.md
```diff
@@ -28,21 +28,22 @@ Therefore, think carefully about the language you use, avoid jargon and subject-
...
-* You only have 280 characters so using hashtags to link to different groups of interest is an important way to get your message out there.
+* Unless you're paying for X Premium you#ll only have 280 characters, so using hashtags to link to different groups of interest is an important way to get your message out there.
 * Tweets with hashtags can increase engagement by almost 100% for individuals.
 * There are also regular hashtags for each day of the week, for example, #MondayMotivation or #FridayFeeling.
 
 **5. Include some other accounts in your tweet**
 
 * Linking your tweet to other accounts using @ will mean you get more retweets and therefore will spread your tweet further.
-* It is a good idea to find other researchers and organisation that have large Twitter followings.
+* It is a good idea to find other researchers and organisation that have large X followings.
```
Arielle-Bennett, line 38:
```suggestion
* It is a good idea to find other researchers and organisations that have large X followings.
```

book/website/communication/social-media/social-media-twitter-tips.md
```diff
@@ -28,21 +28,22 @@ Therefore, think carefully about the language you use, avoid jargon and subject-
...
 This gives your tweet the largest reach possible.
 
 **6. Add visuals**
 
 * You can use images and gifs to draw attention to your tweets.
 * If you don't have an image, a link will generate an image so make sure that the link is up to date and has a relevant image for your tweet.
-* Yow also have the option to add a description of the image so that content is accessible to more people. Here is a link to the twitter help page on [how to make images accessible for people](https://help.twitter.com/en/using-twitter/picture-descriptions).
+* Yow also have the option to add a description of the image so that content is accessible to more people.
```
Arielle-Bennett, line 45:
```suggestion
* You also have the option to add a description of the image so that content is accessible to more people.
```

Arielle-Bennett (commented):
Left some corrections and comments on the number of active users that we quote...

book/website/communication/social-media.md
EstherPlomp: I agree that it would be better to remove it because it will indeed go out of date quickly!

llewelld: I tried to find a more recent number so that this could be updated, but there seems to be some disagreement about whether this number is increasing or decreasing, and X may have [changed how it counts](https://variety.com/2023/digital/news/musk-twitter-x-acquisition-one-year-user-revenue-decline-1235770297/) this recently, so it's all a bit messy.

One reason we might want to keep it is to allow comparison with the other platforms listed on the page, so that their relative sizes can still be compared. Would it make sense to keep it, but making it more vague ("Around 300 million..." or "Several hundred million...")?

book/website/communication/social-media/social-media-twitter-tips.md
llewelld: Thanks for highlighting this one too. I'll update (or remove) this based on what's decided with the other case.

llewelld:
I pushed some changes to add in the LA Times reference alongside the Nature reference (since the Nature article is paywalled, whereas the LA Times one isn't).

book/website/communication/social-media.md
Arielle-Bennett: I'd be interested to hear from others on this (@KirstieJane?) 

On the one hand, I agree it's a useful comparison if we can be confident in the ballpark figure, on the other, the level of uncertainty around it maybe makes it less useful? I'd also argue that in all these cases, the total number of users is less important for researchers than whether their target groups/communities are using the platform, but that's definitely not a _lightly_ opinionated take!

KirstieJane: I have almost no opinion on this but if asked for one I'd say keep the number and put a date after it .

Keeping that number up to date isn't important but some sense of scale is important I think.

llewelld: Thanks for this discussion; this has been really helpful. I've updated the numbers to the latest I could find and added dates into the text so that they're grounded. I've also added in references for them. Maybe these should just be links though?

These numbers came from this [Wikipedia page](https://en.wikipedia.org/w/index.php?title=List_of_social_platforms_with_at_least_100_million_active_users&oldid=1185160092). The references are also from this page; I checked that the links work and state what's claimed.

I'm still more than happy to remove all reference to these numbers if that's the preferred option.

Arielle-Bennett: ah no I have many hills on which I would die, but this is not one of them 😄 let's keep it in and add the dates like you suggested so that it's contextualised for people

book/website/communication/social-media/social-media-twitter-tips.md
llewelld: I've removed this one. On reflection, I agree it's not providing much benefit here.

Arielle-Bennett (approved):
looks good to me @llewelld - thank you for doing this detailed work!

llewelld:
Thank you all for checking the changes and for all of the detailed feedback and discussion!

PR #3418, opened 2023-11-15
Update overview.md - project management framework 

Arielle-Bennett:
I added some of these as review suggestions then realised I should probably just open a PR to commit them instead 😄

-->
### Summary

Adds tags to document, an updated intro and an attempt to get the table to render okay by splitting in 2

- [ ] Everything looks ok?

- [X] All contributors to this pull request are already named in the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file.


kallewesterling (approved):
I don't love splitting up the table like this, but I suppose that if it's the only way to render it decently in JupyterBooks then we'll have to go with it!

Issue #3032, opened 2023-04-06
Adding 2021-2023 TTW Fireside chat organisers/speakers to contributors table


aleesteele:
### Summary

The Fireside Chats are important spaces for The Turing Way community to gather and discuss issues relevant to open science and open ways of working. However, these important community contributions aren't necessary captured in the contributors table.

- Review the co-organisers and speakers at Fireside chats here from this YouTube playlist: https://www.youtube.com/playlist?list=PLBxcQEfGu3DmBcSmt9GVfo4wN1KP1y2yl
- Get the github username of each attendee and add contributors individually
- Add each individual organiser and person with all-contributors bot for 'event-organizing' (don't forget to use the American English spelling 

### Who can help?

* Anyone who has an interest in helping to keep us up to date!

### With the all-contributors bot
Here's an example of how you could get involved with this issue! 

Tag each person's Github username using the syntax below:

`@all-contributors please add @cassgvp for event organising`

Tip: I would recommend adding each person one at a time – as often times, all contributors bot gets a bit confused when more than one PR is open in the project and it causes all sorts of merge conflicts! Unfortunately, this is an ongoing issue & question for the project more broadly (see [here](https://github.com/all-contributors/app/pull/127) and [here](https://github.com/all-contributors/app/issues/146)) -- some folks have struggled with this, as [documented here](https://github.com/all-contributors/all-contributors/issues/375))

### Editing the json file directly

All the contributors & their contributions (which are visibly rendered on our README.md file) are actually stored in this json file: https://github.com/the-turing-way/the-turing-way/blob/main/.all-contributorsrc.

Alternatively, you can make a pull request with a bulk edit of usernames and their contributions

Tip: There might be folks that are already in our all-contributors table, but just need to have another attribute added to their username. Watch out for duplicates!

aleesteele:
1. What exactly is The Turing Way?
- @cassgvp 
- @KirstieJane 
- @malvikasharan

aleesteele:
2. Cross-Community Concerns in Open Communities
- Emmy Tsang - 
- Kate Hertweck
- Tiago Lubiana
- Humberto Julio Debat
- Melissa Mendonça

aleesteele:
3. [Inclusive multilingual data science](https://www.youtube.com/playlist?list=PLBxcQEfGu3DmBcSmt9GVfo4wN1KP1y2yl)
- Batool Almarzouq
- Bobby Shabangu
- Camila Rangel-Smith
[...]

PR #3592, opened 2024-04-02
Update .all-contributorsrc - fixing json

aleesteele:
Fixing JSON file for all contributors bot (found in #3032)


.all-contributorsrc
```diff
@@ -4597,6 +4597,7 @@
       ]
     }
   ],
```
JimMadge, line 4599:
```suggestion
```
This list appears to be ended twice (and this one is too early 🙂)

JimMadge (changes requested):

.all-contributorsrc
aleesteele: Ack, thanks Jim! I ran it through a JSON validator, and it caught the first error and not the second! 😮‍💨

JimMadge (approved):

PR #3617, opened 2024-04-16
Updating 2022 Annual Report Archive

aleesteele:
### Summary
- Updating 2022 report

- Adding in Q1, Q2, Q3, Q4 reports

- [ ] *Lorem ipsum dolor sit amet, consectetur adipiscing.*
- [ ] Everything looks ok?

### DIFF
--- a/project_management/quarterly_reports/2022-turing-tps-report.md
+++ b/project_management/quarterly_reports/2022-turing-tps-report.md
@@ -11,21 +11,11 @@
 - Ethics approval status: Confirmation ethics approval not required
 - Reports drafted by Malvika Sharan
 
-### Overview of the project
+## Overview of the project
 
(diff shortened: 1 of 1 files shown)


malvikasharan:
Thanks, Anne. I have tried to keep other reports as one big report (rather than a quarterly view). I would prefer to keep it that way so we have one single section for each header. That allows a better overview of what we have done in a year. That does make me question, why I still call the report folder quarterly. Let me see if I can help do the merging part here.

aleesteele:
Makes sense! Old quarterly habits die hard! 😁 - and in terms of grabbing stats on a regular basis, we can review how to do that on a more ongoing basis for reporting? Would we want to monitor on a 6 month basis or in a more ongoing way?

PR #3697, opened 2024-06-04
actually distinguish between licence and license

sjvrijn:
### Summary

The Licensing section made a point of distinguishing between American and British English, but for British English, both noun and verb were still spelled as `license`.

- Replace `license` with `licence` when referring to the noun in British English

- [x] The following people should be added to the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file: @sjvrijn

### DIFF
--- a/book/website/reproducible-research/licensing.md
+++ b/book/website/reproducible-research/licensing.md
@@ -19,7 +19,7 @@ No previous knowledge is needed, this chapter explains how important it is to un
 ## Summary
 
 > This chapter was written using American English, in which the word **license** is a noun **_and_** a verb.
-> With British English, however, **license** is a noun (as in, _to issue a license_), while **license** is a verb (as in, _they licensed the event_).
+> With British English, however, **licence** is a noun (as in, _to issue a licence_), while **license** is a verb (as in, _they licensed the event_).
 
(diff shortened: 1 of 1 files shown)


JimMadge (approved):
Good catch, it was exactly the wrong way around :smile:.

RichardJActon (approved):
It's possible this is my fault, I've not checked the history but my spell checking may have done this when I was editing this chapter.

As a dyslexic British person who has strategically adopted the American approach of not distinguishing to reduce cognitive overhead I'm selfishly hoping to contain the scope of this change to this disclaimer and not adopt the distinction across the entire text.

JimMadge:
> As a dyslexic British person who has strategically adopted the American approach of not distinguishing to reduce cognitive overhead I'm selfishly hoping to contain the scope of this change to this disclaimer and not adopt the distinction across the entire text.

That is interesting. I'm normally a bit fussy about American English creeping in to British English. But only using license in software is so common it would only add confusion. I think it is better to stick with the convention that license is a noun and a verb in the OSS world.

JimMadge:
@all-contributors please add @sjvrijn for content

PR #3702, opened 2024-06-04
Add a resource to team manuals (and move one)

EstherPlomp:
### Summary

A small addition of a resource, and moving a resource which got quite some critique due to the length to one of the other examples instead as prominent first example.


Arielle-Bennett (approved):
Looks good to me Esther thank you!

PR #3705, opened 2024-06-04
Glossary: Open Source (Re)Definition

AbcSxyZ:
fix #3653

Proposal for the creation of a new chapter for open source, starting from a blank page to prevent legacy. Introduce a (re)definition of the concept of open source to pave the way for a new understanding of this fundamental notion. The goal is to create a simple page as a basis for future improvements regarding explanations of « open source ».

This is a major modification with deep implications that change the meaning given to open source, in tension with a more traditional sense. Related to [an initiative](https://github.com/the-turing-way/the-turing-way/issues/3653) (#3653) for The Turing Way that raises the question of the meaning of open source around some debates on the confusion surrounding the concept.

Among other questions, there is a debate on whether open source is a concept that applies only to software or if it is an appropriate terminology more broadly for a variety of digital resources whose sources are provided. This confusion over the meaning of open source leads to uncertainty about whether qualifying The Turing Way as open source, as mentioned in its current introduction, is accurate.

## Acknowledgement with proposal validation

This is a deliberately incomplete page with the aim of allowing the community to write what open source is. A validation of the proposal would have multiple significations.

**First, you recognize that The Turing Way is open source,** notably because of the availability of (most) of its sources which could even be an « open source educational resource » more specifically.

The Turing Way ending up being an example of an open source project.

**Secondly, you acknowledge that no current definition of open source is adequate,** with a meaning that is still quite unclear. With a particular mention to the Open Source Definition that can't be used as a reference to define open source as usually because of serious limitations.

This is in line with current debates on the very definition of open source, with interconnected dimensions that make reality difficult to describe. Through this modification, it makes The Turing Way a space where these debates on the definition of open source can take place.

A validation would be a collective acknowledgement that we don't know much about open source and that the concept needs further development.

Is the Turing Way open source ? What is open source ?

## Initial (Rejected) Proposal for Archiving

**[No creation of an open source page with a definition, replaced by an inclusion in the glossary.]**

**Noun:** Practice of providing digital resources with their sources.
**Adjective:** Relating to a digital resource whose source files are available.

> **Examples:**  
>  open source resource, open source educational resource, open source software, open source image, open source hardware…

## Bibliography

A collection of resources on the use of open source beyond software, strictly on open source resources. See also [main comment on the meaning of open source](https://github.com/the-turing-way/the-turing-way/issues/3653#issuecomment-2137470668) in the initial issue of the proposal (#3653).

[...]

PR #3692, opened 2024-06-03
Open Source WG: Specify sources in figures guidelines

AbcSxyZ:
*[In relation to the working group for the 3-7 june book around the question: is The Turing Way Open Source ? (see #3653)]*

fix #3688

Add a section to ask about providing figures sources when adding images around the question of « open source images ».

Remain vague, no guidelines on how/where/whatever (@JimMadge I leave you the fun with the infrastructure WG), in the idea it has to be improved over time.

Use of the qualifier needs, I let you switch to a potential requirement (must) in the future if you consider imposing sources availability. I would love to speak of open source images, I prefer to be patient.

PR #3760, opened 2024-06-12
Homepage: avoid qualifying The Turing Way as open source

AbcSxyZ:
Alternative proposal for #3705 if the concept of « open source » is assumed strictly related to software.

Context for origin of the proposal: https://github.com/the-turing-way/the-turing-way/pull/3705#issuecomment-2161237778

Problematic situation where « open source » is explained and usually assumed in relation to software, but used to qualify the Turing Way handbook in an unclear manner as it is a non-software artefact. This fuels confusion about the concept.

This is coming next to debates on the meaning of open source beyond software, which may be applied to digital resources whose sources are provided. Questioning regarding the fact open source should be used only for software, PR for this case.

Need of an appropriate use of the terminology through the resource and provide explanation that would reflect this understanding.

### DIFF
--- a/book/website/index.md
+++ b/book/website/index.md
@@ -3,7 +3,7 @@
 
 *Welcome to The Turing Way handbook to reproducible, ethical and collaborative data science.*
 
-_The Turing Way_ project is open source, open collaboration, and community-driven.
+_The Turing Way_ project is providing its source files, open collaboration, and community-driven.
 We involve and support a diverse community of contributors to make data science accessible, comprehensible and effective for everyone.
 Our goal is to provide all the information that researchers and data scientists in academia, industry and the public sector need to ensure that the projects they work on are easy to reproduce and reuse.
 
--- a/book/website/index.md
+++ b/book/website/index.md
@@ -3,7 +3,7 @@
 
 *Welcome to The Turing Way handbook to reproducible, ethical and collaborative data science.*
 
-_The Turing Way_ project is providing its source files, open collaboration, and community-driven.
+_The Turing Way_ is an open science, open collaboration, and community-driven project.
 We involve and support a diverse community of contributors to make data science accessible, comprehensible and effective for everyone.
 Our goal is to provide all the information that researchers and data scientists in academia, industry and the public sector need to ensure that the projects they work on are easy to reproduce and reuse.
 
--- a/book/website/reproducible-research/open/open-source.md
+++ b/book/website/reproducible-research/open/open-source.md
@@ -4,7 +4,7 @@
 (rr-open-source-whatis)=
 ## What Is Open Source Software?
 
-When a project is open-source [{term}`def<Open Source Software>`], anybody can view, use, modify, and distribute the project for any purpose.
+When a software is open-source [{term}`def<Open Source Software>`], anybody can view, use, modify, and distribute its source code for any purpose.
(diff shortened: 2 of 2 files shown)


malvikasharan:
Hi Simon, I am not convinced that The Turing Way welcome page needs to be edited the way proposed here. We have agreed that Turing Way is not software, but it is an open project. I'd be convinced to change Open Source to Open Science, just because I am not keen to continue this debate but not the edit you have suggested.

malvikasharan (approved):

book/website/index.md
```diff
@@ -3,7 +3,7 @@
 
 *Welcome to The Turing Way handbook to reproducible, ethical and collaborative data science.*
 
-_The Turing Way_ project is providing its source files, open collaboration, and community-driven.
+_The Turing Way_ is an open science, open collaboration, and community-driven project.
```
malvikasharan, line 6: Hi Simon, I am not convinced that The Turing Way welcome page needs to be edited the way proposed here. We have agreed that Turing Way is not software, but it is an open project. I will add a preferred edit with "open science" (instead of "source", which seems to be the hook here) and merge.

AbcSxyZ:
Wow, I didn't though you would change it that fast 😅 In some extent it wasn't intended to be merged like that, that's why the sentence wasn't quite nice.

Are you sure do that can kind of change like that ? As it (open source) can be part of your speech to present the turing way for some time, I'm surprised. It is also an introduction sentence of the book and also a bit of what led me to do everything I've done here, may be great to discuss it more deeply 😉

And for me it remains pending in relation to the meaning given to the concept through the handbook, as open source seems to be an appropriate qualifier.

It was to open discussions initially. Shouldn't it be reverted ?

P.S.: I could also have thought of open education ? How do you position TTW with it ?

malvikasharan:
My role is to ensure a welcoming community atmosphere and avoid contentious debates over specific wording in the book. The project's main focus is on building and sharing best practices and resources in data science - not defining terminology. While you do continue to discuss the open-source-ness of The Turing Way on pull requests and issues - I want to avoid the disputed terminology. In my talks, I am well-placed to define my point of view.

Next time, if you are not ready for a PR to be merged, please use [DRAFT] in the title, or keep the PR in the draft mode.

AbcSxyZ:
I was thinking mainly of the fact that you use open source a lot in your introductory sentences of the project (your team, not just you) like during presentations. I thought I saw it in an image, but I have some doubts as I can't find it, I don't know if it could be used elsewhere in this way. It was a bit part of the project identity.

I hope my debates don't cause any problems, it's not really a personal desire that the concept is so confusing, it's a pain for me too. This specific word being a fundamental concept in these dynamics of openness, it's a bit annoying.

malvikasharan:
I have suggested edits in #3692, which should address some concerns on the image source file (where they should be posted). I do use "open source" in my talk to describe how the project uses open source infrastructure, builds on open source resources, shares its sources in GitHub etc. It is much easier for me/my team to explain those - compared to edits in this PR where, like you, people who have not previously engaged with the project/community can bring preconceived notions about specific words we use. As said, our goal is to get information easy to access - not to create confusion. If open "source" causes confusions and stops people from reading rest of the sentence/content then I am not going to be stubborn about using that word in the opening line.

PR #3757, opened 2024-06-11
Restructure table of contents

JimMadge:
### Summary

Simplify the table of contents structure by removing the unused parts section.
This shouldn't change the output, but does make the structure of the table of contents YAML easier to read.

### What should a reviewer concentrate their feedback on?

- [ ] Everything looks ok?

### Acknowledging contributors

- [ ] All contributors to this pull request are already named in the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file.
- [ ] The following people should be added to the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file: 

to close #3762

PR #3762, opened 2024-06-12
CI: Update pathways version

bsipocz:
We need a better way for picking up new versions (even just leaving the main branch as the requirement rather than the hash), but this will be good as a hotfix for now.

Required for #3757


JimMadge:
I think this commit won't work because it wants Python more recent than 3.8 (Netlify is making our lives difficult here). I've backported the changes to a 3.8 branch in #3757 .

PR #3784, opened 2024-07-09
Delete newsletter directory (now that it has been transfered to new repo)

da5nsy:
### Summary

As described in https://github.com/the-turing-way/the-turing-way/issues/3352 we're slowly splitting the repo into more manageable chunks. The first chunk is the newsletter folder!

The newsletter directory now has it's own repo at: https://github.com/the-turing-way/newsletter
created as described here: https://github.com/the-turing-way/the-turing-way/issues/3352#issuecomment-1896470970

I'm 90% convinced that this is the right way to go about this, but an alternative proposition would be to not only delete the directory, but to delete it from the history of this repo also. This would have the advantage of reducing the size of the repo.

However, that would:
- rewrite history, which is a pain for projects with lots of open PRs and clones distributed over many people's personal machines
- probably get fiddly because I expect the `/newsletters` directory has been at various different levels/locations/names etc at different times.
- (I also don't technically know how to _do this_ currently, though I expect I could work it out).


bsipocz (approved):

KirstieJane:
This is exciting! Thank you @da5nsy and @bsipocz (and the infrastructure working group!)

JimMadge:
[git filter-repo](https://github.com/newren/git-filter-repo) would probably be the way to go to remove these files from the entire history.

It would mean a force push.

The problem might not be _that_ bad though. Unless someone has modified the newsletter files, there shouldn't be any conflicts. If they have modified the newsletter files, then that should be committed in another repo anyway.

PR #3696, opened 2024-06-04
Update netlify urls

JimMadge:
### Summary

Fixes #3578 

Updates URLs across the repository to use `book.the-turing-way.org` instead of `the-turing-way.netlify.app`.
Also fixes instances of `the-turing-way.netlify.com` (was that an old URL?)

### What should a reviewer concentrate their feedback on?

- [ ] Everything looks ok?

### Acknowledging contributors

- [x] All contributors to this pull request are already named in the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file.
- [ ] The following people should be added to the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file:

PR #4051, opened 2025-02-18
remove multiple dependencies

da5nsy:
## Update: 
Rather than just removing the tornado pin, following [this comment](https://github.com/the-turing-way/the-turing-way/pull/4051#issuecomment-2765702550) this PR has been updated to instead remove most dependencies.

---
### Original text:

`tornado>=6.4.2` is now required by other dependencies, so the manual pin is no longer required.

This time I generated the `requirements.txt` file properly (https://github.com/the-turing-way/the-turing-way/pull/4023#issuecomment-2558127783).

Keeping this as a draft until https://github.com/the-turing-way/the-turing-way/pull/4019 is complete (at which point `requirements.txt` will probably need regenerating).


book/website/requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.13
+# This file is autogenerated by pip-compile with Python 3.12
```
JimMadge, line 2: Let's do this with 3.13 as that is what CI/deployment uses.

JimMadge:
I think we can _probably_ remove all dependencies apart from `jupyter-book` and `pathways` now.

da5nsy:
> I think we can _probably_ remove all dependencies apart from `jupyter-book` and `pathways` now.

I'll try it out.

da5nsy:
I've tested this locally and nothing obvious breaks.

For future reference:
```
python3.13 -m venv ./venv
source ./venv/bin/activate
python -m pip install --upgrade pip
cd book
pip install -r ./website/requirements.in
make pathways
make serve
pip install pip-tools
cd website
pip-compile requirements.in
```

da5nsy:
I'm not sure I understand @JimMadge... `pip-compile requirements.in` generates a new `requirements.txt` file, which will _then_ be used in CI and Netlify (and for other local builders). I only used `pip install -r ./website/requirements.in` because I was modifying `requirements.in` in this PR.

JimMadge:
> I'm not sure I understand @JimMadge... `pip-compile requirements.in` generates a new `requirements.txt` file, which will _then_ be used in CI and Netlify (and for other local builders). I only used `pip install -r ./website/requirements.in` because I was modifying `requirements.in` in this PR.

What were you trying to do? I don't think you should need to install the `.in` file at all.

da5nsy:
Re: plausible, it looks like @JimMadge removed it [here](https://github.com/the-turing-way/the-turing-way/commit/ff736987ce37c5cb18453a0947aeb6b193bac2a0) but there's no linked discussion in https://github.com/the-turing-way/the-turing-way/pull/3696 so I'm wondering if it got killed by accident?

We _do_ have google analytics as of [this commit](https://github.com/the-turing-way/the-turing-way/commit/cb2a9a9c1b55ed19a17a773af5f26246b231a7d9#diff-210ce24ae4d5231e886d770bbdb3020ff3f944b931194eda685037a9d8c57f3a) by @malvikasharan.

JimMadge:
> When I do this I only get mainly trivial diffs to yours @JimMadge, with the only interesting one being that I do not have a requirement for appnope whereas it looks like you do? It looks like [ipykernal's appnope dependency is system specific](https://github.com/ipython/ipykernel/blob/v6.29.5/pyproject.toml#L33C14-L33C39) (I hadn't seen this before).

Yeah looks like that is the case, I ran `pip-compile` on Darwin 😆. You can remove it if you like, I don't think the runners should need it.

da5nsy:
Just for interest - I was curious what would happen to the mac-os CI build with `appnope` removed from `requirements.txt` and it looks like it does recognise it as a missing dependency and add it back in, specifically for the macos build.
The Windows build, meanwhile, adds in `colorama`, `pywin32`, and `pywinpty`.

I think this is probably the behaviour that we want, but it slightly updates my understanding that the packages in `requirements.txt` are the _only_ packages that are used in a build - it's them _plus_ anything that is a dependency that wasn't listed (and this is prone to pop up if the CI is running on a different OS than the one used to generate the `requirements.txt` file).

JimMadge:
@da5nsy Yeah that makes sense to me.

`pip install -r requirements.txt` will install packages according to the spec in that file, and pip automatically fetches dependencies. So, for the MacOS build those dependencies are fetched.


The trick is, using `pip-compile` will resolve a full cascade of dependencies, your direct dependencies and all sub-dependencies. But, because of platform/arch differences that full list could be different depending on where you run `pip-compile`. `pip-sync` (also in the `pip-tools` package) will give you _exactly_ what is in a `requirements.txt`.

PR #4057, opened 2025-02-18
Add `--strict` to CI

da5nsy:
We merged #4019 when it was good enough, and it seems like we might be waiting for some upstream fixes (to DOI handling), but ideally we would add the `--strict` flag back to the CI eventually.

This PR is to keep track of where that is up to, and hopefully chip away at any warnings/errors.


da5nsy:
Crumbs: https://github.com/the-turing-way/the-turing-way/pull/4019#discussion_r1958857992

da5nsy:
This is now just waiting on the wikipedia link bug mentioned [here](https://github.com/jupyter-book/mystmd/issues/1336#issuecomment-2593403732).

Issue made to track: https://github.com/jupyter-book/mystmd/issues/1954

JimMadge:
@da5nsy thinking about this. If `--strict` returns 0 when there are still warnings, I think that isn't what we want, or at least it is a regression. Before we failed on any warnings, which I think is important for maintaining quality and warnings can indicate things like missing figures and references.

da5nsy:
@JimMadge your diligence conflicts with my desire to tick off this PR, but I think you're probably right 😭

However, I don't know how much of this current behaviour is informed by: https://github.com/the-turing-way/the-turing-way/blob/eb1f678739509bdb132f1c67d7bb47c3baf0be81/book/website/myst.yml#L3-L5

We should test it without that. If I get chance I'll report back here. 

Depending on results we might want to report upstream.

da5nsy:
> Before we failed on any warnings, which I think is important for maintaining quality and warnings can indicate things like missing figures and references.

We should probably also do a test of introducing some common problems and see whether they get elevated to errors currently. 

To test:
- [ ] Page missing from contents 
- [ ] Duplicated figure reference
- [ ] Missing bib entry
- [ ] Refer to a section which doesn't exist
- [ ] See https://github.com/the-turing-way/the-turing-way/pull/4287/files for ideas
- [ ] _Any other common ones we can foresee?_

JimMadge:
I think we probably _do_ want to ignore the link check, assuming that also tests external links. Sometimes those will fail for reasons outside of our control and I wouldn't want that to block people.

book/Makefile
```diff
@@ -20,7 +20,7 @@ strict:
 	cd website && jupyter-book build --html --strict
 
 ci:
-	cd website && jupyter-book build --html --ci
+	cd website && jupyter-book build --html --ci --strict
```
da5nsy, line 23:
```suggestion
	cd website && jupyter-book build --html --ci --strict --link-checker=false
```
not sure if this is the right format, let's see...

da5nsy:
I've asked on discord if there is a way to turn off the external link check but keep the internal link check with warnings raised to errors

da5nsy:
> I've asked on discord if there is a way to turn off the external link check but keep the internal link check with warnings raised to errors

I've made a GH issue here: https://github.com/jupyter-book/mystmd/issues/2122

da5nsy:
> > Before we failed on any warnings, which I think is important for maintaining quality and warnings can indicate things like missing figures and references.
> 
> We should probably also do a test of introducing some common problems and see whether they get elevated to errors currently.
> 
> To test:
> 
>     * [ ]  Page missing from contents
> 
>     * [ ]  Duplicated figure reference
> 
>     * [ ]  Missing bib entry
> 
>     * [ ]  _Any other common ones we can foresee?_

Testing on top of the latest commit (which is completing locally for me, and completed on Netlify, but struggling in github CI - lack of caching link-check? Or the fact that we're running link-check 4 times simultaneously and so hitting limits?)

## Page missing from contents
Comes up as an error in main log but not at summary (which is weird)

## Duplicated figure reference
(I just duplicated a whole figure)
Comes up as warning in main log and summary, but not raised to error.

## Missing bib entry
This seems to have had zero effect. Further investigation required.

PR #4058, opened 2025-02-18
Add licences to myst.yml

da5nsy:
This should add little buttons to each page that link to the licences we use.

Taken from [here](https://github.com/jupyter-book/mystmd/blob/0ca2c6792b3f83d58942788cf499d495a9e3e73d/docs/myst.yml#L5-L7).


JimMadge (approved):
Ooo, very nice. Looks great in the preview 👍.

PR #4093, opened 2025-03-28
Missed a few edits in updating governance docs

malvikasharan:
@Arielle-Bennett made some edits which seem to have not been integrated.

This PR reintroduces them.

A few future PRs will be created when we transition our infra (emails for example) out - which is not in this PR!


Arielle-Bennett (approved):
Thanks for catching these @malvikasharan 🌷

PR #4103, opened 2025-03-31
align local build instructions

da5nsy:
We currently have local build instructions in 2 separate places. This PR removes one and replaces it with a link to the other.


JimMadge (approved):

JimMadge:
I think, let's only have the instructions in one place. And in this project that place feels like it should be the book :+1:.

PR #4267, opened 2025-06-19
Add windows-specific command for local build

da5nsy:
`source ./venv/bin/activate` does not work in Windows.
An alternative which works for me is `venv\Scripts\activate` (found [here](https://stackoverflow.com/a/61679638/6464224)).

Adding this as a tab set in case there are other OS-specific commands we want to include later. 
Currently everything else works on Windows for me - though, I suspect I have installed some extras (e.g. `make`) previously and so we might want to try and run a build on a fresh install sometime and see if/where it breaks.

### DIFF
--- a/book/website/community-handbook/local-build.md
+++ b/book/website/community-handbook/local-build.md
@@ -56,7 +56,7 @@ $ python3 -m venv ./venv
 Next, active the virtual environment,
 ::::{tab-set}
 :::{tab-item} Unix
-:sync: linux
+:sync: Unix
 ```console
 $ source ./venv/bin/activate
 ```


book/website/community-handbook/local-build.md
```diff
@@ -54,10 +54,21 @@ $ python3 -m venv ./venv
 ```
 
 Next, active the virtual environment,
-
+::::{tab-set}
+:::{tab-item} Linux
```
sgibson91: Maybe call this Unix so it covers both Linux and Mac since the same command works on both? (Nitpicky, I admit!)

sgibson91 (approved):

book/website/community-handbook/local-build.md
da5nsy:
```suggestion
:::{tab-item} Unix
```

book/website/community-handbook/local-build.md
```diff
@@ -54,10 +54,21 @@ $ python3 -m venv ./venv
 ```
 
 Next, active the virtual environment,
-
+::::{tab-set}
+:::{tab-item} Unix
+:sync: linux
```
da5nsy:
```suggestion
:sync: Unix
```

PR #4156, opened 2025-05-07
Enable Cross-Reference in MyST via CORS

rowanc1:
This PR updates the Turing Way’s site configuration to allow **Cross-Origin Resource Sharing (CORS)** from all origins (`Access-Control-Allow-Origin: *`). This change is being made to:

* Support **MyST cross-references** (e.g., `xref` links, [see docs](https://mystmd.org/guide/external-references#myst-xref)) **from other sites** that use The Turing Way as a citation source.
* Enable content embedding, linking, or automated metadata access (e.g., for previews or API consumers) **without authentication or same-origin constraints**.

#### What is CORS and why is it needed?

**CORS** is a browser security feature that restricts how resources on a web page can be requested from another domain outside the one that served the web page. For example, JavaScript running on `external-site.org` cannot fetch metadata or assets from `https://book.the-turing-way.org/` unless explicitly allowed.

By setting CORS headers to allow all origins (`*`), we make it possible for external tools and sites to:

* Reference Turing Way content directly via [structured and stable links](https://mystmd.org/guide/external-references#myst-xref).
* Preview or embed sections of the book from that page (with attribution).
* Use it in federated, cross-site knowledge systems (like MyST Markdown references in external books or educational hubs).

Since the Turing Way is **public and does not require authentication**, allowing all origins does **not pose a security risk** and aligns with the project's goals of openness, reusability, and interoperability. This is also the default on GitHub pages as a comparison.

PR #4279, opened 2025-06-26
remove CI path filtering

da5nsy:
### Summary

@the-turing-way/infrastructure-working-group recently turned on branch protection rules for `main`, meaning that certain checks would need to pass before merging was allowed.

This had the unintended consequence of blocking certain PRs from being merged (by anyone other than a user with admin permissions who could bypass the branch protection rules). This affected PRs which **only** modified files which were outside of the CI path spec:

https://github.com/the-turing-way/the-turing-way/blob/226f9b2ace7c8dcaccf104999e0cda9eaf836b63/.github/workflows/ci.yml#L11-L16

For example: 
- all-contributors-bot PRs, e.g. https://github.com/the-turing-way/the-turing-way/pull/4276
- things which just change netlify deploy info: https://github.com/the-turing-way/the-turing-way/pull/4156
- others (?)

This is because of [the way github had chosen to implement path filtering](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs):

> If a workflow is skipped due to [path filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), [branch filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore) or a commit message (see below), then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.

[Previous docs](https://github.com/github/docs/blob/285e455bf41892520fdf2503384bb47cb48e53dd/content/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks.md?plain=1#L58) also said:

> For this reason you should not use path or branch filtering to skip workflow runs if the workflow is required.

Here I remove the path filtering.

We could hypothetically add in some pseudo-path filtering, using [conditions](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution), which is what was recommended in [historic github docs](https://stackoverflow.com/a/78003720/6464224), but I propose not doing that, since our current build time is only ~1 minute. This means that avoiding unnecessary builds is a low priority compared to readability of the CI workflow script.

Any issues I may not have anticipated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,12 +8,6 @@ on:
   pull_request:
     branches:
       - main
-    paths:
-      - "book/**"
-      - ".github/workflows/**"
-      - "ACCESSIBILITY.md"
-      - "CONTRIBUTING.md"
-      - "contributors.md"
 
 # set the default shell to bash for all jobs
 # this is important for the windows OS!
--- a/.github/workflows/lorem-ipsums.yml
+++ b/.github/workflows/lorem-ipsums.yml
@@ -12,8 +12,6 @@ on:
   pull_request:
     branches:
       - main
-    paths:
-      - "book/**"
 
 # Set up the Continuous Integration job
 jobs:
--- a/.github/workflows/no-bad-latin.yml
+++ b/.github/workflows/no-bad-latin.yml
@@ -12,8 +12,6 @@ on:
   pull_request:
     branches:
       - main
-    paths:
-      - "book/**"
 
 # Set up the Continuous Integration job
 jobs:


KirstieJane:
I'm not going to review because I don't feel qualified to approve BUT this does seems super sensible to me!

Thank you @da5nsy and @sgibson91 for all your help and expertise 🙏 🌟

da5nsy:
Adding a link to previous discussions of protections: https://github.com/the-turing-way/the-turing-way/pull/2016

JimMadge (approved):
Looks sensible.

It feels wasteful to sometimes run CI when you don't need to. However, this has caused problems with the required checks, so I think this is right.

What might be cool is if we had an early exit. Like

```
if not_modified([book/**, ...]) {
  return 0
}
```

Not sure if that is possible in workflows.

PR #3651, opened 2024-05-13
Code style additions

AoifeHughes:
### Summary

Addresses #3650 and also adds information following on from #1390 

This PR adds a new section on pre-commit hooks to the "Code Style and Formatting" chapter. 

It is also important to decide if the `.pre-commit` file added should be ran. If so then can we do it here (~500 files will be changed by having line-endings etc. corrected)

### List of changes proposed in this PR (pull-request)

* Added a new section titled "Pre-commit hooks" to the "Code Style and Formatting" chapter
* Introduced the concept of pre-commit, its key features, and how to set it up in a project
* Placed the new section between the "Automatic formatting" and "Online services providing software quality checks" sections

### What should a reviewer concentrate their feedback on?

- [ ] Does the new section fit well within the overall structure of the chapter?
- [ ] Is the information on pre-commit presented clearly and accurately?
- [ ] Are there any additional details or examples that should be included?

### Acknowledging contributors

- [ ] The following people should be added to the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file: @AoifeHughes

PR #4285, opened 2025-07-03
fix: remove BOM in file

agoose77:
### Summary

@stefanv was looking at The Turing Way website, and noticed that a particular file heading doesn't behave properly:

screenshot of the reproducible research resources page - the markdown title is showing but so too is the label text that should be hidden in the html render

After a little bit of digging, I found a BOM at the head of the file, which is invisible to the human eye but not to the tokeniser:
```
$ cat -v /book/website/reproducible-research/overview/overview-resources.md| head
M-oM-;M-?# Resources

For additional resources like videos and reference papers on reproducibility, see the {ref}`rr-overview-resources-reading` and {ref}`rr-overview-resources-addmaterial` sections.

(rr-overview-resources-learn)=

## What to Learn Next?
```

This PR strips that mark. I searched for other files, and this is the only one affected.

> [!Note]
> Should we introduce a Git hook (pre-commit) to catch this? e.g. https://gist.github.com/rlee287/e6026243dedc38398298

### DIFF
--- a/book/website/reproducible-research/overview/overview-resources.md
+++ b/book/website/reproducible-research/overview/overview-resources.md
@@ -1,4 +1,4 @@
-﻿(rr-overview-resources)=
+(rr-overview-resources)=
 # Resources
 For additional resources like videos and reference papers on reproducibility, see the {ref}`rr-overview-resources-reading` and {ref}`rr-overview-resources-addmaterial` sections.
 


KirstieJane:
I added some alt text to the screenshot in the text 😄

KirstieJane:
This is such a random find! Thank you so much @stefanv and @agoose77 for tracking it down!

KirstieJane:
I'm going to leave the question about the test to @the-turing-way/infrastructure-working-group team!

JimMadge:
I've seen some things which may be related. I think sometimes these odd characters and whitespace find their way in when people copy-paste from a rich text editor.

We did have [an attempt to make the style more consistent with pre-commit](https://github.com/the-turing-way/the-turing-way/pull/3651). I'm pretty supportive of that. I think there was a concern about it being confusing or a barrier to contribution. (Say for instance, pre-commit blocking someone from making a commit, or a bot adding new commits to your branch).

We can open a discussion if anyone is interested.

JimMadge:
@all-contributors please add @stefanv for bug

Issue #4177, opened 2025-05-19
Deploy Portuguese book


JimMadge:
### Context

From [a conversation](https://github.com/the-turing-way/the-turing-way/issues/3255#issuecomment-2886731537) regarding #3255.

There is a strong and active community contributing in Portuguese.
If we add a Portuguese deployment we will,

- Have the value of a deployment in one more language
- Have an easier entry to learning about supporting multiple languages
  (without having to develop new tools or worry about supporting an arbitrary number of languages from the beginning)

Could be a good collaborative piece of work for @the-turing-way/translation and @the-turing-way/infrastructure-working-group 

### Action Items

- [x] Move up the `images` subdirectory one level so it doesn't need to be duplicated for each language subfolder
- [x] Move up non-language specific files such as the requirements files
- [x] Copy Portuguese content into `book/pt` subfolder of this repo
- [x] Update the Makefile to permit building both languages, either in a "build all" manner or a `make build pt` option way
- [x] Update CI workflow to build and test Portuguese site
- [ ] Update Netlify config to host another site at book.the-turing-way.org/pt (There may be some prior art in the Turing Way Translation team regarding the Arabic version)
- [ ] Work with Translation team to update Crowdin workflow to pull changes into this repo

### Definition of Done

- [ ] The deployment of the current site and all its features continues to work
- [ ] There is a live deployment of the Portuguese translations at book.the-turing-way.org/pt
- [ ] (stretch) The Crowdin workflow for the Portuguese translation flows back into _this_ repo

### Nice-to-haves, but out of scope for now

- Language switcher, for the MVP we will work with URL manipulation
- Full-blown PR change analysis to only rebuild websites who's content/config has changed, a la https://www.youtube.com/watch?v=b7CCUHYOaK4

sgibson91:
@BatoolMM if you're going to be at the next Collab Cafe (whenever it is! 😂 ) I'd love to chat to you about this and refine this issue a bit more to have some action items

sgibson91:
I did some digging around the multilingual examples Danny found in #3255. While it's great stuff, I'm concerned that the example site is JB1 using a sphinx extension, and we are running JB2 which I don't think it sphinx-compatible anymore. So we may have to replicate the TeachBooks' folks extension as a JavaScript plugin that generates myst AST, which will be a significant chunk of work.

sgibson91:
I updated the issue body with some action items and a definition of done based on my investigations and discussions with the team today.

[...]

PR #4289, opened 2025-07-07
Deploy Portuguese book

da5nsy:
**Replaced by https://github.com/the-turing-way/the-turing-way/pull/4290**

---

Partially implements #4177 (which in turn partially implements #3255).

Specifically:

- [ ] Copy Portuguese content into `book/pt` subfolder of this repo
- [x] Update the Makefile to permit building both languages, either in a "build all" manner or a `make build pt` option way
- [x] Update CI workflow to build and test Portuguese site
- [ ] Update Netlify config to host another site at book.the-turing-way.org/pt (There may be some prior art in the Turing Way Translation team regarding the Arabic version)
- [ ] Work with Translation team to update Crowdin workflow to pull changes into this repo


da5nsy:
Note: the changed files in 915c9b4783bc461356cb1e379e07e1fa864e8af0 are those which are likely to need some re-working longterm to make work multi-lingually - so far I've just switched out `website` for `en`

da5nsy:
I'm not happy with way I've done the Makefile currently:
- `make serve` won't work - I think that would need to be language specific (although maybe whatever solution we find for combining the output files for Netlify could be mirrored here).
- I'd prefer it be work in a "for language in languages do x" way rather than a new line for each but: 
   - a) I couldn't get `jupyter book build` to accept a path as input, so I stuck with the `cd` approach 
   - b) I couldn't get Makefile loops to work for me last night

 Let's discuss at this afternoon's @the-turing-way/infrastructure-working-group meeting?

@BatoolMM - if you are able to and fancy joining this afternoon's meeting that might be particularly handy (no worries if not!)

da5nsy:
@the-turing-way/infrastructure-working-group (I think @JimMadge in particular) - do you have feelings on `book.the-turing-way.org/pt` vs `pt.book.the-turing-way.org`?
I quite like the latter because it means that the right side of the URL is content and the left is meta, but I don't know if there's technical limitations/preferences. 

Particularly thinking about redirects @JimMadge (from book.TTW to book.TTW/en or en.book.TTW, at least as a temporary thing while we transition from English-first to multi-lingual)

JimMadge:
I would prefer `book.the-turing-way.org/<lang>/`. It feels more conventional and we could support that with [redirects based on the client's region or language](https://docs.netlify.com/routing/redirects/redirect-options/#redirect-by-country-or-language).

It would be a big pain to have to configure DNS for every language.

`<lang>.book.the-turing-way.org` feels like we are running a different web server for each language, which I don't think is the idea, right?

JimMadge:
We will probably want to use [ISO 639 two letter language codes](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes#Table), perhaps with extensions for regions(?).

book/Makefile
```diff
@@ -15,20 +15,24 @@ upgrade:
...
 ci:
-	cd website && jupyter-book build --html --ci
+	cd en && jupyter-book build --html --ci
+	cd pt && jupyter-book build --html --ci
 
 # Clean any build outputs and artifacts
 clean:
-	jupyter-book clean --all --yes ./website/_build
+	jupyter-book clean --all --yes ./en/_build
+	jupyter-book clean --all --yes ./pt/_build
```
sgibson91:
```suggestion
	jupyter-book clean --all --yes ./**/_build
```
Would this not work?

JimMadge:
Probably reaching the limit of what is sensible to do with a Makefile. We can find an alternative in the future when we are happy with the other aspects of deploying multiple languages though.

book/Makefile
da5nsy: Yes... sort of.

For reasons unclear to me I had to run it twice:
```
(venv) danny@hades:~/Documents/the-turing-way/book$ make clean
jupyter-book clean --all --yes ./**/_build
⛔️ pt/_build Error reading file: Error: EISDIR: illegal operation on a directory, read
⛔️ website/_build Error reading file: Error: EISDIR: illegal operation on a directory, read
Deleting all the following paths:

  - pt/_build/cache
  - pt/_build/html
  - pt/_build/templates
  - website/_build/cache
  - website/_build/html
  - website/_build/site
  - website/_build/templates

🗑  Deleting: pt/_build/cache
🗑  Deleting: pt/_build/html
🗑  Deleting: pt/_build/templates
🗑  Deleting: website/_build/cache
🗑  Deleting: website/_build/html
🗑  Deleting: website/_build/site
🗑  Deleting: website/_build/templates
(venv) danny@hades:~/Documents/the-turing-way/book$ make clean
jupyter-book clean --all --yes ./**/_build
⛔️ pt/_build Error reading file: Error: EISDIR: illegal operation on a directory, read
Deleting all the following paths:

  - pt/_build/site

🗑  Deleting: pt/_build/site
```

but I've noticed some weird behavior with `make clean` before (generally that EISDIR thing) so I suspect that's indpendent of this way of doing it.

da5nsy: Used here: https://github.com/the-turing-way/the-turing-way/pull/4290/commits/9831f0cfab92fe74f4ff4811fc56673486ebdac0

da5nsy:
Closing in favor of a blank slate at https://github.com/the-turing-way/the-turing-way/pull/4290

PR #4263, opened 2025-06-18
Replace pathways with curation plugin

JimMadge:
### Summary

Replaces the pathways script with a MyST plugin.

This has a number of advantages,

- Single build command (there is no pathways build now, pathways are always included)
- More maintainable plugin
- Possibility for others to use the more generic curation directive
- Update pathways sections in the community handbook
  - Guide for authors
  - Information for infrastructure WG

Closes https://github.com/the-turing-way/pathways/issues/20

### List of changes proposed in this PR (pull-request)

* Remove pathways dependency
* Add curations plugin
* Add curation pages for pathways

### What should a reviewer concentrate their feedback on?

- [ ] Everything looks ok?

### Acknowledging contributors

- [ ] All contributors to this pull request are already named in the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file.
- [ ] The following people should be added to the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file:

PR #4294, opened 2025-07-10
Update Netlify docs to remove pathways reference

da5nsy:
Following #4263

### DIFF
--- a/book/README.md
+++ b/book/README.md
@@ -29,7 +29,7 @@ _The Turing Way_ book is built and deployed online using [Netlify](https://www.n
 If you want to deploy the book on Netlify, you'll need the following settings:
 
 - Base directory: `book/`
-- Build command: `make deps && make pathways && make ci`
+- Build command: `make deps && make ci`
 - Publish directory: `website/_build/html`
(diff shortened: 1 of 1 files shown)


sgibson91 (approved):

book/README.md
```diff
@@ -29,7 +29,7 @@ _The Turing Way_ book is built and deployed online using [Netlify](https://www.n
 If you want to deploy the book on Netlify, you'll need the following settings:
 
 - Base directory: `book/`
-- Build command: `make deps && make pathways && make ci`
+- Build command: `make deps && make ci`
 - Publish directory: `website/_build/html`
```
JimMadge, line 33: I'm not really sure what this is supposed to mean.

I don't like duplicating the build command here. It would be better just to reference `netlify.toml` here.

JimMadge (changes requested):

book/README.md
da5nsy: Agreed. I'll merge this (since it's currently incorrect) and then make a new issue for later.

PR #4303, opened 2025-07-16
Update book directory README

### LINKED ISSUE #4296: reference `netlify.toml` in `book/readme` rather than reproducing the build command
### Summary

See https://github.com/the-turing-way/the-turing-way/pull/4294#discussion_r2200113526

### What needs to be done?

_No response_

### Who can help?

_No response_

JimMadge:
Closes #4296

### DIFF
--- a/book/README.md
+++ b/book/README.md
@@ -8,11 +8,7 @@ All figures associated to the chapters are stored in, and linked from, the [`fig
 
 ### Configuration
 
-- The table of contents (TOC) defines the order of chapters as they appear in the book.
-To change the TOC, please edit the `website/myst.yml` file with correct information on filenames and their relative locations in this repository.
-Documentation on controlling the TOC structure can be found on the [Jupyter Book website](https://jupyterbook.org/customize/toc.html).
-- Same applies for more general configuration using `website/myst.yml`.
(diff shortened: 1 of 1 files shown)


book/README.md
```diff
@@ -1,38 +1,18 @@
...
 
-### Deploying
+### Building
 
-The site is built automatically using the two directories `book` and `book/website`.
-All of the dependencies are specified in `dependencies.in`, which is transformed into the lock file `dependencies.lock` using `pip-compile`.
-
-Instructions for how to build a Turing Way book locally can be found in [_The Turing Way_'s Community Handbook](https://book.the-turing-way.org/community-handbook/local-build).
-
-#### On Netlify
```
da5nsy, line 25: I think this might be the only place we document our use of Netlify besides redirects page, and the toml file itself. 
This probably isn't the best place for it, but we might want to keep it in until we're replacing it with something else - what do you think @JimMadge?

da5nsy (approved):
LGTM!

book/README.md
JimMadge: It is the wrong place and would be forgotten about (again 😆).

I don't think the information here is very useful so would say we should wait until we know there is a need for information before writing it.

Issue #4056, opened 2025-02-18
Jupyter Book 2 migration


JimMadge:
- [x] https://github.com/the-turing-way/the-turing-way/pull/4019
- [x] https://github.com/the-turing-way/the-turing-way/pull/4053
- [ ] https://github.com/the-turing-way/the-turing-way/issues/4055

sgibson91:
> * Convert old style links _e.g._ `[](foo)` to `[](#foo)` ([@sgibson91](https://github.com/sgibson91) I'm not sure I understand this one)

Ok, so we'd create references like `(hello-world)=` and would reference them as `[](hello-world)`, but now they need to be referenced as `[](#hello-world)` instead. I _think_ I got all the ones that at least raised warnings/errors in CI in my PR.

sgibson91:
We may also want to revert [this commit](https://github.com/the-turing-way/the-turing-way/pull/4019/commits/b3dc4b073b256156cccbd17944c363fbb7377e0c) which makes checking links less strict, simply because there were a TONNE of external links that were 404'ing or not resolving that I didn't know what to do with (e.g., things like collaborators personal websites, and I think a Turing Institute course got moved or made private?)

malvikasharan:
Dear Infra team, we are working on a few community projects using the Pathway feature. Would you be able to share tentatively a timeline that I can convey to them for the feature being live again? For clarity, there is no hurry.

JimMadge:
No plan yet. I was thinking about having a look to try and understand what needs to be done in the collab café today though.

PR #4315, opened 2025-07-29
GitHub user role

JimMadge:
Contributes to #4056 and #4057

Adds a plugin with a role to create links to a GitHub profile from a username.
Fixes inconsistent way GitHub usernames are presented in the contributors record, and unescaped @ characters being interpreted as citations.

Longer term, we really should make a contributor directive, there is a huge amount of manual formatting in the contributors record which is very messy and inconsistent.


da5nsy (approved):
Very cool!

Minor points:

## Documentation

Should we document this somewhere? Here perhaps? - community-handbook/infrastructure/infrastructure-contributors#personal-highlights
Perhaps a link to `book/website/plugins/github.mjs`

And could we put a link inside `book/website/plugins/github.mjs` to the myst(?) documentation that describes how one would make/edit one of these?

## Using in bullets but not in main text

Looks like you've chosen to not use the mechanism in main text areas ([e.g.](https://github.com/the-turing-way/the-turing-way/pull/4315/files#diff-c0b113c623775f02613d14453acd67f658c62b9a1255190a3caf35bb6fb32c21R277)) but only in the bullet lists, is this a technical limitation or a choice?

JimMadge:
> Should we document this somewhere? Here perhaps? - community-handbook/infrastructure/infrastructure-contributors#personal-highlights Perhaps a link to `book/website/plugins/github.mjs`
> 
> And could we put a link inside `book/website/plugins/github.mjs` to the myst(?) documentation that describes how one would make/edit one of these?

Yes, definitely should do that. I think there are two types of documentation,

1. How to use this (in the style guide)
2. How this is implemented (in the infrastructure section)

In both cases I think it makes sense to create a "custom roles and directives" page and put it there (assuming we will have more in the future).

> Looks like you've chosen to not use the mechanism in main text areas ([e.g.](https://github.com/the-turing-way/the-turing-way/pull/4315/files#diff-c0b113c623775f02613d14453acd67f658c62b9a1255190a3caf35bb6fb32c21R277)) but only in the bullet lists, is this a technical limitation or a choice?

Not sure I understand, but are you asking can you use the role outside of an unordered list, like

```Markdown
Just a paragraph {githubuser}`UserName` with the role in
```

If so, yes. We can use it wherever we write MyST.

da5nsy:
> Not sure I understand, but are you asking can you use the role outside of an unordered list

Yep. I note you have used it in situations like [this](https://github.com/the-turing-way/the-turing-way/pull/4315/files#diff-c0b113c623775f02613d14453acd67f658c62b9a1255190a3caf35bb6fb32c21R270), but not [this](https://github.com/the-turing-way/the-turing-way/pull/4315/files#diff-c0b113c623775f02613d14453acd67f658c62b9a1255190a3caf35bb6fb32c21R277), and was wondering why.

JimMadge:
@da5nsy I've added some documentation.

da5nsy:
Great, thanks Jim.

If we were to add a line to the documentation which was something like "the advantage of using a directive rather than a basic markdown link is...", how would that line end? (I think I'm not sure).

JimMadge:
I've added a short explanation.

da5nsy:
Thanks Jim!

da5nsy (approved):

PR #4323, opened 2025-07-30
Use token in pull request action

JimMadge:
Uses a PAT token, `CROWDIN_ACTION_TOKEN`, in the pull request action in the Crowdin workflow.

A token with [contents:write and pull-requests:write](https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#token) is needed when assigning teams as reviewers.

This seems to be working now.
https://github.com/the-turing-way/the-turing-way/actions/runs/16619792625


.github/workflows/crowdin-contributors.yml
```diff
@@ -44,6 +44,7 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v7
         with:
+          token: ${{ secrets.CROWDIN_ACTION_TOKEN }}
```
sgibson91:
```suggestion
          token: ${{ secrets.GITHUB_TOKEN }}
```

[This block](https://github.com/the-turing-way/the-turing-way/pull/4323/files#diff-264b5bf640da600c6608ec79c90839d8fe4fd6d7f39c916517e8a10a8539defbL14-L16) sets the proper permissions on `GITHUB_TOKEN` (for this workflow only), but it wasn't being passed to the action. If we do it this way, we won't have to manage a secret in the repo.

sgibson91 (changes requested):

sgibson91:
I also recommend removing this section and the associated environment: https://github.com/the-turing-way/the-turing-way/pull/4323/files#diff-264b5bf640da600c6608ec79c90839d8fe4fd6d7f39c916517e8a10a8539defbL18-L20

.github/workflows/crowdin-contributors.yml
JimMadge: Oh, so it does. My reading of the docs made me think the runner token wasn't sufficient.

Much better than having to generate and manager the PAT ourselves :tada:.

JimMadge:
> I also recommend removing this section and the associated environment: https://github.com/the-turing-way/the-turing-way/pull/4323/files#diff-264b5bf640da600c6608ec79c90839d8fe4fd6d7f39c916517e8a10a8539defbL18-L20

I _think_ that is required for the Crowdin action, this secret is in that environment

https://github.com/the-turing-way/the-turing-way/blob/b4fa3f04c031a292a79dd6dada7dc02a0613baba/.github/workflows/crowdin-contributors.yml#L41

sgibson91:
Yes, you're right. I just had the same thought and was double checking myself

.github/workflows/crowdin-contributors.yml
sgibson91: Probably with the default permissions, the runner token _isn't_ sufficient, so the permissions block lets you tweak on a per job/per workflow basis

sgibson91 (approved):

JimMadge:
@sgibson91 looks to be failing after that change maybe the PAT is required? https://github.com/the-turing-way/the-turing-way/actions/runs/16623050715/job/47032028659

Perhaps it is to do with the scope of the token rather than the permissions?

sgibson91:
Just trying one more thing...

sgibson91:
Yep, you're right:

> Error: Unable to request reviewers. If requesting team reviewers a 'repo' scoped PAT is required.

Damn it, would've worked if we weren't also trying to request reviewers 😆 

Ok, go back to the original version of your PR, but can we add the crowdin PAT to the existing crowdin environment in the repo settings? That way, it's only available to workflows that call that environment.

sgibson91:
If adding the PAT, the `permissions` block can also just be removed

PR #4324, opened 2025-07-30
Remove unused workflow

sgibson91:
This has been disabled for a long while


sgibson91:
I have also deleted the associated GitHub Actions environment/secret

da5nsy:
Ah, shame, I always hoped we'd get this working. Maybe we still will one day!

sgibson91:
No reason the PR can't be reverted! But also, I think getting images out of the repo altogether is the better goal.

PR #4327, opened 2025-07-31
Fix bib item's author field

JimMadge:


EstherPlomp (approved):
Thank you for fixing! I'll hopefully remember that for future bib updates... :)

PR #4329, opened 2025-07-31
Fix German README

stweil:
### Summary

The German README had some issues which are fixed by this pull request.

- [ ] The following people should be added to the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file: <!-- replace this text with the GitHub IDs of any new contributors


HeidiSeibold:
It all looks good to me :) 🎉 

I personally would use "Du" instead of "Sie" as that is quite formal. But that's personal preference and I think it can be merged and discussion on how formal we want to be left to another time.

HeidiSeibold (approved):

stweil:
I kept "Sie" because it was already used in the text, but I agree that using "Du" could be considered after a discussion.

PR #4287, opened 2025-07-06
Formatting and broken link fixes

da5nsy:
Pulling just the formatting fixes out of https://github.com/the-turing-way/the-turing-way/pull/4057, because https://github.com/the-turing-way/the-turing-way/pull/4057 is blocked by https://github.com/jupyter-book/mystmd/issues/2122 IMO, but the bug fixes include quite a few broken internal links which I think would be good to merge asap.

Partially addresses https://github.com/the-turing-way/the-turing-way/issues/4266

PR #4337, opened 2025-08-14
make new strict sandbox branch

da5nsy:
Picking up from https://github.com/the-turing-way/the-turing-way/pull/4057 (which was inadvertently merged, because I created https://github.com/the-turing-way/the-turing-way/pull/4287 as a branch from it)


da5nsy:
Moving to #4406

PR #4358, opened 2025-09-16
Fix external link check job

### LINKED ISSUE #4145: [BUG]: External link check workflow broken
### Description

The [external link check workflow](https://github.com/the-turing-way/the-turing-way/blob/main/.github/workflows/external_link_check.yml) has been [failing for the last ~6 months](https://github.com/the-turing-way/the-turing-way/actions/workflows/external_link_check.yml), meaning that [this issue](https://github.com/the-turing-way/the-turing-way/issues/3171) hasn't been getting updated.

We do however have 2 other running link checkers currently, and so we might want to have a wider think about which one we should prioritise (especially if fixing this one isn't a quick fix).
- [The one which runs on every PR within CI](https://github.com/the-turing-way/the-turing-way/blob/c0778e4787d4b2fc7089553a4ff13e63d489454c/.github/workflows/ci.yml#L78-L105) (but doesn't seem to actually report to anywhere??)
- the one built into JB2 which we're [currently ignoring](https://github.com/the-turing-way/the-turing-way/blob/c0778e4787d4b2fc7089553a4ff13e63d489454c/book/website/myst.yml#L3-L5)

JimMadge:
### Summary

Fixes #4145 

* Stops step failure upon lychee finding errors
* Correct root-dir for offline link check

- [ ] Everything looks ok?

### DIFF
--- a/.github/workflows/external_link_check.yml
+++ b/.github/workflows/external_link_check.yml
@@ -37,7 +37,7 @@ jobs:
         uses: lycheeverse/lychee-action@v2
         with:
           args: --config './lychee.toml' --no-progress '--root-dir "$GITHUB_WORKSPACE/book/website/_build/html/" book/website/_build/html/**/*.html'
-          fails: false
+          fail: false
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}} # Use runner's GitHub token to avoid rate limiting
 


da5nsy:
I'm struggling to remember why we needed the `error_rules` config (https://github.com/the-turing-way/the-turing-way/pull/4358/commits/950cd49335376427793ebe77a3a3bb25a18ece66) but if it's building fine without it then let's trim it and see if anything breaks without it. I suspect the things it was letting pass have been resolved [tentative celebration 🎉 ]

.github/workflows/external_link_check.yml
```diff
@@ -34,9 +34,10 @@ jobs:
           restore-keys: cache-lychee- # restore latest cache (not a specific commit)
 
       - name: Check links
-        uses: lycheeverse/lychee-action@v2.0.2
+        uses: lycheeverse/lychee-action@v2
         with:
           args: --config './lychee.toml' --no-progress 'book/website/_build/html/**/*.html'
+          fails: false
```
da5nsy: I'd have assumed we'd want it to fail on error, no?

https://github.com/lycheeverse/lychee-action?tab=readme-ov-file#passing-arguments

sgibson91:
I brought in the error rules config when I swapped to JB2, because there are a load of links in the book that don't go anywhere any more (like individual's websites, I think the Turing Institute made some repos/resources internal as well) and at the time it was too much to go through every single one and extract them all. It also wasn't obvious with what to replace them either.

.github/workflows/external_link_check.yml
sgibson91: If you fail at this point, the whole workflow will stop and the subsequent steps won't execute. Since they are the steps that create the issue with the report, we want that behaviour.

JimMadge:
> I brought in the error rules config when I swapped to JB2, because there are a load of links in the book that don't go anywhere any more (like individual's websites, I think the Turing Institute made some repos/resources internal as well) and at the time it was too much to go through every single one and extract them all. It also wasn't obvious with what to replace them either.

Were you running build with `--strict` or `--check-links`? When I build on this branch, it doesn't seem to check links (at least there are not errors/warnings). So I think we are safe to remove it. Might have been a change of behaviour in MyST.

sgibson91:
I can't remember. Possibly strict because we ended up removing that as well

JimMadge:
> I can't remember. Possibly strict because we ended up removing that as well

In that case, we might want to reintroduce that config section in #4337

da5nsy:
Looks like there's still an error unfortunately: https://github.com/the-turing-way/the-turing-way/actions/runs/17790875886/job/50585370684#step:7:127

JimMadge:
Ah, I didn't look closely enough at the logs 🤦

PR #4406, opened 2025-10-31
Run strict build on Ubuntu build only

sgibson91:
### Summary

Run the strict build on Ubuntu only.

Hopefully by not doing this on all 3 OSs, we won't cause GitHub Actions to timeout

- [ ] Everything looks ok?


.github/workflows/ci.yml
```diff
@@ -38,8 +38,14 @@ jobs:
         working-directory: ./book
 
       - name: Build Jupyter-Book with warnings raised to errors
-        run: make ci
         working-directory: ./book
+        run: |
+          if [[ "${{ matrix.os }}" == *"ubuntu"* ]]; then
+            echo "Running strict build on Ubuntu"
+            make ci --strict
```
sgibson91:
```suggestion
            make strict
```

JimMadge:
Well, it doesn't time out, but it doesn't pass 😆.

I think we need to get the changes that let us selectively ignore broken links in, then we should be close to enabling `--strict` in CI.

sgibson91:
Yeah, that's why it's draft! I was experimenting 😁

JimMadge:
Hopefully with the 2.0.0 release, this should work now :eyes:.

Well, apart from the _other_ warnings 😆

JimMadge:
Hmm, it isn't. I'm not sure why, they `key` argument seems to break ignoring `link-resolves`. If you remove the key all link errors are ignored.

I can't reproduce this in a toy example though 🫠

JimMadge:
I think the answer was, the glob pattern wasn't what we wanted. It uses [picomatch](https://www.npmjs.com/package/picomatch#globbing-features) where `*` is a glob _excluding_ `/`.

JimMadge:
> Site has 3 errors and 4 warnings, stopping build.

This feels very doable :rocket:

da5nsy:
"All checks have passed" 😱 ✨

JimMadge:
Oh, interesting. There are still some warnings at the top, mainly invalid YAML blocks in directives, and some notes about implicit references. I suppose those don't cause a strict build to fail 🤔.

JimMadge (approved):

PR #4407, opened 2025-10-31
Add helper redirects for the community calendar

JimMadge:
This PR adds two redirects for our calendars,

- The web view of the calendar
- The ics file for subscribing to the calendar

I've added DNS rules and domain alias in Netlify, so hopefully this should work (pending DNS rules propagating) 🤞


da5nsy (approved):
LGTM!
Nice idea!

Issue #4398, opened 2025-10-29
Simplify glossary reference style


JimMadge:
### Summary

Currently, the guidance for referring to a glossary item, for example 'Binderhub' is to write,

```
Binderhub [{term}def<Binderhub>]
```

which will render like Binderhub \[[def](./)\].

This is complicated to write, probably prone to error and presumably inconvenient for screen readers.

A much simpler recommendation would be to write,

```
{term}`Binderhub`
```

which will render like [Binderhub](./)

or, if you want to change how the term is written,

```
{term}`mybinder.org <Binderhub>`
```

which will render like [mybinder.org](./).

### What needs to be done?

- [ ] Update the guidance in the community handbook `style-glossary.md`
- [ ] Update instances of the old syntax

### Who can help?

- Anyone who is confident with find/replace or a tool like vimgrep

Arun24-8:
i want to solve this problem

PR #4409, opened 2025-11-04
Simplify glossary reference syntax (#4398)

Arun24-8:
Replaced deprecated [{term}`def<Term>`] syntax with the simpler {term}`Term` style,
as per the updated glossary style guide.


JimMadge (commented):
Thanks @Arun24-8 :tada:.

I'm sure there are many other examples of the old syntax in the book. Would you like to try and fix more or all of them?

The other task is to update the guidance in `style-glossary.md` which says to use the more complicated style.

Arun24-8:
"No, I don’t want to update them right now."
Thanks for asking me @JimMadge

JimMadge:
OK, I think it would be simpler to fix all instances in one PR, updating the community handbook at the same time, so I will close this.

PR #4410, opened 2025-11-04
Fix(website): Correct SVG display in dark mode

### LINKED ISSUE #4120: [BUG]: SVG corners getting cut off on `https://book.the-turing-way.org/afterword/afterword`
### Description

In dark mode the image on https://book.the-turing-way.org/afterword/afterword looks like this:

Screenshot showing dark bottom corners

### Your personal set up

- **Operating System**: Ubuntu 24.04.2 LTS
- **Browser**: Firefox 137.0

Vamsiindugu:
### Summary

Fixes #4120

This PR addresses the issue where SVG image on the afterword page had their corners cut off when viewed in dark mode. This was due to a reference issue that has now been resolved.

**Before:**

Screenshot of the original image, showing the bottom corners are cut off in dark mode

**After:**

Screenshot of the corrected image, showing the bottom corners are now displayed correctly

* Modified `book/website/afterword/afterword.md` to correct the reference of image.

**How to test:**
1.  Navigate to the afterword page: https://book.the-turing-way.org/afterword/afterword
2.  Switch to dark mode.
3.  Verify that the image is displayed correctly without any cutoff corners.

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -778,7 +778,8 @@ about open and reproducible (data) science. I am currently thinking a lot about
 how we could use the content from The Turing Way and turn it into a course.
 This idea was also part of an application, where I proposed to start a new
 group on *Open AI*.
-> Specifically, I have co-authored these chapters: [Research Compendia](https://book.the-turing-way.org/reproducible-research/compendia.html), [File Naming Convention](https://book.the-turing-way.org/project-design/filenaming.html), and reviewed many contributions.
+> Specifically, I have co-authored these chapters: [Research Compendia](https://book.the-turing-way.org/reproducible-research/compendia.html), [File Naming Convention](https://book.the-turing-way.org/project-design/info-management/filenaming/
+), and reviewed many contributions.
 > I regularly recommend The Turing Way as a resource. Both for learning more
 about reproducible data science and also when discussing specific topics.
 I think that people are taking it on and reading it :)
@@ -1052,7 +1053,7 @@ I am also an associate member of the Chartered Institute of Personnel and Develo
 
 * Personal highlights:
 >I started familiarizing myself with the GitHub environment and learned how to collaborate through it. It was especially nice to meet the vibrant community working together on the Turing Way!
-> Specifically, I have authored a subchapter on [Research Data Management Toolkit](https://book.the-turing-way.org/reproducible-research/rdm/rdm-toolkits.html).
+> Specifically, I have authored a subchapter on [Research Data Management Toolkit](https://book.the-turing-way.org/reproducible-research/rdm/rdm-resources/#research-data-management-toolkits).
 
 
 <!---L--->


da5nsy (changes requested):
Hi @Vamsiindugu. I'm afraid we still want to use the svg version (or the appropriate file format for the chosen output - that's why we use the `.*`) rather than hard-coding the `.jpg` version. 

I think this might be an upstream issue with svg rendering, either in Jupyter Book or the browser.

Vamsiindugu:
Hi @da5nsy,

Thank you for the feedback and for pointing me in the right direction.

I've done a deeper investigation based on your comments. You are correct that hard-coding the `.jpg` is not the right fix. However, I've also confirmed that this is **not a CSS or styling bug** in dark mode.

The issue is with the **original SVG file itself**. I've opened the file, and the graphic is simply missing its bottom corners.

Here is the image in the directory [the-turing-way/book/figures](https://github.com/the-turing-way/the-turing-way/blob/main/book/figures/road-to-reproducibility.svg):

Screenshot of the original SVG on GitHub, showing the bottom corners are missing

Here is the updated SVG image I have. It has the correct corners and I'd like to use it as the replacement:

The corrected 'road to reproducibility' SVG image, now with complete bottom corners

Would it be acceptable for me to update this PR by replacing the old, problematic SVG file with the new, correct one?

Please let me know if that's the right way to proceed, and I'll push the changes.

Thanks!

da5nsy:
Thanks @Vamsiindugu - can you provide some more details on how you modified the svg please?

Vamsiindugu:
Yeah @da5nsy I used the .jpg image to compare with .svg and came to know that both are same then after confirming that I headed to recraft.ai website and uploaded the .jpg file to convert into .svg file. I exported that file and replaced that old svg with a new one.

da5nsy:
Hi @Vamsiindugu, the svg was the original file produced by the artist and is of higher resolution (not quite the right terminology I think, because it is a vector image, but hopefully you know what I mean). 
If we are to modify the svg file here then I would want to look into which specific lines of the svg need to be changed to remove this artifact, rather than creating a new svg.
More fundamentally, I think we should understand the problem and resolve it, and not just work around it.
This issue probably applies to lots of other images in the repo, and perhaps in other repos, and perhaps in images which will be created at future events by the same artists, and so I think it is worth understanding the root issue.

JimMadge:
Agreed with @da5nsy. Editing the SVG to add a background _should_ be easy and retains the illustration.
Inkscape is probably a good tool to do this, or imagemagick.

da5nsy:
Taking a code approach to it - it looks like replacing [this line](https://github.com/the-turing-way/the-turing-way/blob/97d3f1734402fc131bd42734cff5177171c665e3/book/figures/road-to-reproducibility.svg?short_path=23d0adb#L198) with `<rect width="100%" height="100%" fill="#f0f0f0"/>` resolves the issue.

It looks like the background was set by adding a big ellipse, and since this was probably done on a system which wasn't using dark mode (and so the background behind the background was white...) it wasn't noticed that it didn't totally cover the canvas.

---

I worked this out by opening the image in Inkscape and seeing that it overflowed the canvas in a big ellipse, then looking at the elements in https://www.svgviewer.dev/ and then asking Qwen3-235B how to add a background to an svg file.

da5nsy:
It looks like it would also be possible to resolve this issue with some custom css which forces the background to all svgs to be white - https://github.com/the-turing-way/the-turing-way/pull/4411
I like this approach because it resolves it for all other images without us having to go through each one to check whether it is affected.
The downside is complexity bloat on the book (it would be nice not to have to introduce custom CSS).

What do you think @Vamsiindugu and @JimMadge?

JimMadge:
I think custom CSS is fine. It feels like an odd solution to me though. We've fixed a problem with one image by modifying _all_ svgs. It might be confusing or cause unexpected problems, like if we ever have an svg with an intentionally transparent background.

da5nsy:
> I think custom CSS is fine. It feels like an odd solution to me though. We've fixed a problem with one image by modifying _all_ svgs. It might be confusing or cause unexpected problems, like if we ever have an svg with an intentionally transparent background.

I agree - I think on balance I prefer just editting this specific svg as described [here](https://github.com/the-turing-way/the-turing-way/pull/4410#issuecomment-3488626066) and not using the custom CSS.
I do vaguely recall this being an issue with other svgs too, but we can cross that bridge if/when we come to it.
I was mainly just interested in whether _there is_ a way to do this with CSS 🤓 😅

da5nsy:
@Vamsiindugu - if that all sounds good to you, would you like to modify this PR branch to just do a line edit on the original SVG?

Vamsiindugu:
Hi @da5nsy and @JimMadge,

Thank you so much for the detailed guidance! I've followed your recommendation.

I have reset the branch to be in sync with `main` and pushed a single new commit. This commit applies the line edit to the correct `road-to-reproducibility.svg` file, replacing the `<ellipse>` with the `<rect>` as you suggested.

This should resolve the dark mode issue without replacing the file or adding custom CSS. Please let me know if this looks correct!

da5nsy:
> it looks like replacing [this line](https://github.com/the-turing-way/the-turing-way/blob/97d3f1734402fc131bd42734cff5177171c665e3/book/figures/road-to-reproducibility.svg?short_path=23d0adb#L198) with <rect width="100%" height="100%" fill="#f0f0f0"/> resolves the issue.

My mistake, it should've been `ffffff`. Fixed now.

Vamsiindugu:
Hi @da5nsy,

Thank you for applying the final color fix!

This PR should now fully resolve the SVG display issue. With your approval and all checks passing, it looks complete from my side.

If no further changes are needed, please feel free to merge.

JimMadge:
Bypassing merge requirements as the problem was probably due to rate limiting.

JimMadge:
@all-contributors please add @Vamsiindugu for content and design

PR #4443, opened 2025-11-11
Enforce capitalisation for titles

### LINKED ISSUE #4374: Add CSS style sheet for consistent titles
Maybe we can use a [style sheet](https://mystmd.org/guide/website-style#style-sheet) and [text-transform](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform) to make sure this happens everywhere?

_Originally posted by @JimMadge in https://github.com/the-turing-way/the-turing-way/pull/4373#pullrequestreview-3332228923_

For example,

```css
h1, h2, h3, h4, h5, h6 {
    text-transform: capitalize;
}
```

JimMadge:
### Summary

Fixes #4374 

Styles headings on pages and scrollbars to capitalise all words.
This makes the titles appear consistent and removes the burden from writers to use a consistent case.

* Add style sheet and use
* Apply `text-transform: capitalize` to
  * All heading levels
  * `primary-scrollbar` - the left hand navigation
  * `CollapsibleContent` - the right hand in-page navigation

- [ ] Everything looks ok?

The `CollapsibleContent` class sounds broad, but I couldn't see it being used elsewhere.

### DIFF
--- a/book/website/style.css
+++ b/book/website/style.css
@@ -2,3 +2,12 @@
 h1, h2, h3, h4, h5, h6 {
     text-transform: capitalize;
 }
+
+/* Capitalise headings in sidebar navigation elements */
+.primary-scrollbar {
+    text-transform: capitalize;
+}
+
+.CollapsibleContent {
+    text-transform: capitalize;
+}


da5nsy:
With my infra hat on - I approve. All LGTM.
With my reader hat on - I think I don't like it. Looking at titles which are in captitalised rather than title-case ("Presenting At Remote Events" rather than "Presenting at Remote Events" feels jarring to me. I don't know how much of this is an uncommon personal preference of mine though.

sgibson91:
@da5nsy I would also be in camp "`the, in, or, at` etc shouldn't be capitalised unless they start the heading". But I don't know how easy that is to implement in this scenario though, and wouldn't advocate for spending loads of time on it.

Also, how does this behave? Will it cause a failed build if the style sheet is violated, or is it more pre-commit-like and will auto-correct?

JimMadge:
I would prefer title case too, but that isn't an option for `text-transform`. So, it is this or chaos 😆.

It affects how the browser styles the content, so it doesn't matter what case the source is in and won't cause any CI problems.

da5nsy:
Yeah I did a bit of a dive into it and the internet told me pretty conprehensively that you can't do titlecase in CSS 😢 
I guess, if I'm being blunt, I personally prefer the human-decided chaos. I find the result more readable. But I won't block if there's a general shared opinion that capitalised is preferred over the current situation.

da5nsy:
>  

1. Yes. It me.
2. Why did the alt-text bot not shout at you!? I feel short-changed because I wanted to see what on earth you had written for that and there was but and empty void 😢  (CC: https://github.com/the-turing-way/the-turing-way/pull/4446) 🤔

sgibson91:
> 2\. Why did the alt-text bot not shout at you!?

Screenshot 2025-11-11 at 18 56 33

Because technically, there is alt text.

(P.S. I'm gonna get shouted at for this one. You're welcome.)

(P.P.S. In the theme of human-decided chaos, we are now wildly off-topic.)

KirstieJane:
Came here to vote for human generated chaos over awkward automated capitalisation.

Not a blocking vote, just an opinion!

JimMadge:
Feels like the consensus here is to not make this change?

da5nsy:
> Feels like the consensus here is to not make this change?

I think let's give it a bit of time and see what other folks think. I feel a bit sad for us to not merge it because you've done some lovely work in finding a very nice technical solution here 🥲

EstherPlomp:
I think I agree with @da5nsy about the readability but I also don't really have a strong opinion so I'm happy to go whichever road you/infrastructure team prefers!

Arielle-Bennett:
I have conflicting opinions! 
* It's nice to reduce the burden on maintainers to remember to enforce title capitalisation and other formatting standards we'd like to have consistently across the book. I'm less worried about losing the human chaos as I think there is still plenty of human chaos in the guides and community 😁 
on the other hand.... 
* I too dislike the loss of title capitalisation - having all caps feels off to my copy editor brain 😭

sgibson91:
Compromise by only applying this to h1 instead of all headings?

LizHareDogs:
I'm on the fence about lot like Arielle. 

- Things that facilitate consistency are good. In my own reports I'm never sure what to do with capitalization after the #s.
- But it does really seem like we should be using title case!

JimMadge:
To be clear, the options `text-transform` gives are to capitalise _all_ words, upper case, and lower case. There is no option for title case

So, I think the choice here is between capitalisation or no standardisation.

jezcope:
Another edge case that I don't know if we care about but thought I would mention: some words (often brand names but sometimes other things) are canonically [UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case), which would also be lost with this change.

An alternative approach that is softer but more flexible could be a CI check (like the Lorem Ipsum and Latin Phrases checks) that doesn't block deployment but flags up any inconsistencies, which could then either be fixed or added to an ignorelist on a case-by-case basis.

JimMadge:
> An alternative approach that is softer but more flexible could be a CI check (like the Lorem Ipsum and Latin Phrases checks) that doesn't block deployment but flags up any inconsistencies, which could then either be fixed or added to an ignorelist on a case-by-case basis.

I think I don't like this idea. If we don't block then it will tend to be ignored, and it could get very noisy.

Maybe if we did it based on the diff rather than the whole book, so it would only flag new titles.

JimMadge:
I think there isn't appetite for this, so I'm going to close.

PR #4456, opened 2025-11-11
Provide Green CI and feedback when not building the book for all-contributors PRs

sgibson91:
### Summary

- **The problem:** We don't want to build the book for all-contributors bots, but we want to require the build to pass for other PRs. Filtering for this behaviour is difficult using default GitHub Actions behaviour.

* The original `if` statement is moved from the build job-level, to all the original steps in the job
* Add new steps:
  * Extract the number of the PR
  * Post a comment on the PR explaining what has happened, why, and what the human reviewer should check.

- [ ] Everything looks ok?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,10 +12,6 @@ defaults:
   run:
     shell: bash
 
-permissions:
-  issues: write
-  pull-requests: write
-
 jobs:
   build-jb:
     name: Build
@@ -26,6 +22,9 @@ jobs:
       fail-fast: false # Don't cancel all jobs if one fails
 
     runs-on: ${{ matrix.os }}
+    permissions:
+      issues: write
+      pull-requests: write
 
     steps:
       - name: Extract the PR number from the ref


.github/workflows/ci.yml
```diff
@@ -12,11 +12,13 @@ defaults:
   run:
     shell: bash
 
+permissions:
+  issues: write
+  pull-requests: write
```
JimMadge: Would `pull-requests: write` mean the action can push commits to a PR?

I don't think these are particularly scary, but we might want to move the jobs which don't require this into a different workflow to avoid giving permissions where they aren't needed.

JimMadge (approved):
Looks good.

A little messy but I don't know a better way around that, like an early exit from a job.

JimMadge:
@the-turing-way/infrastructure-working-group I think I must have made a mistake in writing the `CODEOWNERS` file. I think the team should have been tagged for this PR.

.github/workflows/ci.yml
sgibson91: I'll move the block to the job level, then we don't have to split the file

sgibson91:
I'll merge this tonight when I'm not also doing the day job so I can react if it doesn't work as I expect

.github/workflows/ci.yml
JimMadge: Even better

PR #4484, opened 2025-11-22
docs: replace HackMD references with Framapad in templates.md

abhiramch018:
Updated all HackMD references to Framapad in templates.md 
as The Turing Way has moved from HackMD to Framapad.


JimMadge:
The templates are still (mostly) Markdown, so I don't think these changes make sense.

PR #4501, opened 2025-11-28
Fix case in glossary

JimMadge:
When referencing a term in the glossary, for example ``{term}`issue` ``, the case matches what is in the glossary.

Currently, glossary items are usually capitalised, so most often you need to write ``{term}`issue <issue>` `` to avoid capitalised words appearing in the middle of a sentence.

This PR fixes the case for glossary terms, making them lowercase except for proper nouns and acronyms/initialisms.

When using the term role, the argument is not case sensitive so existing uses still work.


Arielle-Bennett (approved):

Arielle-Bennett:
thanks to noticing and tackling this @JimMadge !

PR #4529, opened 2026-02-10
upgrade jb

da5nsy:
### Summary

Upgrades jupyter book to latest version, supplanting https://github.com/the-turing-way/the-turing-way/pull/4528.

## Steps followed:
```
python3.13 -m venv ./venv
source ./venv/bin/activate
pip install pip-tools
cd book
[manually update requirements.in]
pip install -r ./requirements.in
pip-compile --upgrade
pip-sync
git switch -c upgradejb
git add .
git commit -m "upgrade jb"
git push --set-upstream origin upgradejb
```

### DIFF
--- a/book/website/project-design/pd-overview/pd-checklist.md
+++ b/book/website/project-design/pd-overview/pd-checklist.md
@@ -11,7 +11,7 @@ Finally, we can plan for publishing and sharing research components before, duri
 ## Aims & Values 
 * Define the main research questions and objectives.
 * Identify the core values and principles that guide your project.
-* Useful documentation: [project canvas](https://canvanizer.com/new/project-canvas), [values document](https://rebelsguidetopm.com/what-are-your-project-values/), [project 1-pagers](https://www.smartsheet.com/content/project-report-templates).
+* Useful documentation: [project canvas](https://canvanizer.com/new/project-canvas), [values document](https://web.archive.org/web/20240523192550/https://rebelsguidetopm.com/what-are-your-project-values/), [project 1-pagers](https://www.smartsheet.com/content/project-report-templates).
 
 ## Timeline & Milestones 
 * Establish a project timeline with key milestones, output, and deadlines.


JimMadge:
> Can we turn off dependabot trying to change the lockfile? (I mean that other pr also shows why we should not can this requirements.txt but a lockfile.).

We haven't configured dependabot to update package version. That PR was a security update. We can disable that too if we want to. Although, then I think we would also loose it for GitHub actions.

bsipocz:
> We haven't configured dependabot to update package version. That PR was a security update. We can disable that too if we want to. Although, then I think we would also loose it for GitHub actions.

I think we can keep it for GHA, I really don't see how else we would update those version. But if the monthly is too frequent, switch it to quarterly.

OTOH, I think we need to turn it off to do any kind of update for the lockfile. We don't really distribute anything, so I fail to see how any security incidents that doesn't result in a yanking of a version from pypi would be serious enough to tinker with that file.

JimMadge:
> > We haven't configured dependabot to update package version. That PR was a security update. We can disable that too if we want to. Although, then I think we would also loose it for GitHub actions.
> 
> I think we can keep it for GHA, I really don't see how else we would update those version. But if the monthly is too frequent, switch it to quarterly.
> 
> OTOH, I think we need to turn it off to do any kind of update for the lockfile. We don't really distribute anything, so I fail to see how any security incidents that doesn't result in a yanking of a version from pypi would be serious enough to tinker with that file.

There are two distinct types of dependabot update, #4528 was a security update, which could happen for pip or Actions. But that is different from the monthly GitHub actions check for updates. The security PRs occur based on CVEs being reported. So in that case, we don't have control over when they happen.

I think the advantage of adding a dependabot item for pip packages is it would prompt us to update JB/myst without having to manually run `pip-compile`. Like in this PR, we missed a minor version bump that fixed #4246. I'd be quite happy with a monthly PR to make sure we don't fall too far behind.

bsipocz:
> There are two distinct types of dependabot update, https://github.com/the-turing-way/the-turing-way/pull/4528 was a security update, which could happen for pip or Actions. But that is different from the monthly GitHub actions check for updates. The security PRs occur based on CVEs being reported. So in that case, we don't have control over when they happen.

Yeap, I know. But maybe this shows that we should put the indirect dependencies into a different file as those should not be updated by dependabot, neither dependabot.

> I think the advantage of adding a dependabot item for pip packages is it would prompt us to update JB/myst without having to manually run pip-compile. Like in this PR, we missed a minor version bump that fixed https://github.com/the-turing-way/the-turing-way/issues/4246. I'd be quite happy with a monthly PR to make sure we don't fall too far behind.

I would be 👍 for adding the direct dependencies to a pip dependabot. I fully followed the JB release that fixed that bug report, but totally forgot to bump the version here.

PR #4572, opened 2026-03-08
Update sensitive data

EstherPlomp:
update text/internal linking in sensitive data chapters

### Checklist

- [x] I agree to follow _The Turing Way_'s [code of conduct](https://book.the-turing-way.org/community-handbook/coc/)
- [x] I have read the [contribution guide](https://book.the-turing-way.org/community-handbook/contributing/)
- [x] I have read the [style guide](https://book.the-turing-way.org/community-handbook/style/)

### Summary

addresses #4560 
Fixes #4319
Fixes #4442

### List of changes proposed in this PR (pull-request)

* add a couple of resources
* update internal linking and referencing

### What should a reviewer concentrate their feedback on?

- [ ] Everything looks ok?

### Acknowledging contributors

- [x] All contributors to this pull request are already named in the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file.
- [ ] The following people should be added to the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file:

PR #4570, opened 2026-03-05
feat: Document teams as code management methodology

sgibson91:
### Checklist

- [x] I agree to follow _The Turing Way_'s [code of conduct](https://book.the-turing-way.org/community-handbook/coc/)
- [x] I have read the [contribution guide](https://book.the-turing-way.org/community-handbook/contributing/)
- [x] I have read the [style guide](https://book.the-turing-way.org/community-handbook/style/)

### Summary

- fixes #4530 

### List of changes proposed in this PR (pull-request)

* Adds short documentation on the teams-as-code repo in the infrastructure section of the Community Handbook

### What should a reviewer concentrate their feedback on?

- [ ] Any information missing?
- [ ] Everything looks ok?

### Acknowledging contributors

- [x] All contributors to this pull request are already named in the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file.
- [ ] The following people should be added to the [table of contributors](https://github.com/the-turing-way/the-turing-way/blob/main/README.md#contributors) in the README file:

PR #4598, opened 2026-04-01
Fix set of warnings

JimMadge:
### Checklist

- [x] I agree to follow _The Turing Way_'s [code of conduct](https://book.the-turing-way.org/community-handbook/coc/)
- [x] I have read the [contribution guide](https://book.the-turing-way.org/community-handbook/contributing/)
- [x] I have read the [style guide](https://book.the-turing-way.org/community-handbook/style/)

### Summary

Tidies some warnings introduced in #4572 and #4570.
These are difficult to catch if we don't build locally as they won't cause CI to fail. (It feels like there is momentum behind the idea of removing link checks from `--strict` so hopefully we can enable that soon).

Maybe we should do something to reinforce that cross-references should be prefixed with `#` now?

- [ ] Everything looks ok?


EstherPlomp (approved):
Arggh sorry @JimMadge - I thought I was paying attention to the #'s in the internal references but clearly not enough :( 

Thank you thank you!

JimMadge:
> Arggh sorry @JimMadge - I thought I was paying attention to the #'s in the internal references but clearly not enough :(
> 
> Thank you thank you!

It's fine. I think the real problem is our CI doesn't give feedback for those things. There are some fairly serious problems that result in warnings, like missing figures or broken references.

PR #4619, opened 2026-05-12
add `--strict` flag to CI build

da5nsy:
Follows:
- https://github.com/the-turing-way/the-turing-way/pull/4337
- https://github.com/the-turing-way/the-turing-way/pull/4057
- https://github.com/the-turing-way/the-turing-way/pull/4406
- https://github.com/the-turing-way/the-turing-way/pull/4428
- ...
- https://github.com/jupyter-book/mystmd/pull/2819 :sparkles:


da5nsy:
Re: https://github.com/the-turing-way/the-turing-way/pull/4057#issuecomment-2792435696

> > Before we failed on any warnings, which I think is important for maintaining quality and warnings can indicate things like missing figures and references.
> 
> We should probably also do a test of introducing some common problems and see whether they get elevated to errors currently.
> 
> To test:
> 
>     * [ ]  Page missing from contents
> 
>     * [ ]  Duplicated figure reference
> 
>     * [ ]  Missing bib entry
> 
>     * [ ]  Refer to a section which doesn't exist
> 
>     * [ ]  See https://github.com/the-turing-way/the-turing-way/pull/4287/files for ideas
> 
>     * [ ]  _Any other common ones we can foresee?_

---

++ Intentionally broken DOI

da5nsy:
Seeing some residual issues, reported to mystmd folks here: https://github.com/jupyter-book/mystmd/pull/2819#issuecomment-4432569516

book/website/myst.yml
```diff
@@ -18,6 +18,10 @@ project:
       severity: ignore
       key:
         - '{http,https}://**'
```
JimMadge: Now that there is a different flag for link checking, can we remove this rule?

book/website/myst.yml
```diff
@@ -18,6 +18,10 @@ project:
...
+    - rule: doi-link-valid
+      severity: ignore
+      key:
+        - '**'
```
JimMadge, line 28: Which DOIs were giving us problems? It would be good to have this check if we can, to prevent bad DOIs being commited.

JimMadge (commented):
I think I missed the change in behaviour upstream. Looks like we can get this in soon.

book/website/myst.yml
bsipocz: in my recollection it's not specific DOIs but sometimes we just overwhelm the system. I wonder if we could instead build in some timelag checking too many links serially? (if yes, it's a feature for upstream)

da5nsy:
> I think I missed the change in behaviour upstream. Looks like we can get this in soon.

Yep!

I think we just need to do some intentional breaking to test that it fails when we'd want it to, then we're good to go!

book/website/myst.yml
da5nsy: @JimMadge - info here: https://github.com/jupyter-book/mystmd/pull/2819#issuecomment-4432569516

book/website/myst.yml
da5nsy: Yeah I think so

book/website/myst.yml
JimMadge: Yeah that is an example with a valid DOI failing. So I would say in CI disabling this rule is the best approach.

JimMadge (commented):
The changes here look good to me.

`--strict` does summarise warnings (like incorrect cross-reference labels). It doesn't seem to return a non-zero exit code on warnings though. Not sure if that is intended behaviour,

```
 --strict                Summarize build warnings and exit non-zero on
                          any errors. (default: false)
```

@agoose77 is that expected?

bsipocz:
> It doesn't seem to return a non-zero exit code on warnings though.

Yes, I believe it is, I needed to modify a couple of codes to be errors rather than warnings. I would certainly be one who could/would prefer to have a sphinx `-W` like behaviour that non-zero exits on warnings.

JimMadge:
> > It doesn't seem to return a non-zero exit code on warnings though.
> 
> Yes, I believe it is, I needed to modify a couple of codes to be errors rather than warnings. I would certainly be one who could/would prefer to have a sphinx `-W` like behaviour that non-zero exits on warnings.

Oh, I think I see. We can change `severity` to `error` where we want. https://mystmd.org/guide/settings#setting-error-rules

I think having broken cross-references as errors is sensible for The Turing Way. That is what we had before with JB 1.

JimMadge (approved):
I've updated the rules in `myst.yml` to raise some warnings to errors. Have a look at those and say if you think that is the right set.

bsipocz (approved):
The changes to `error_rules` look reasonable

PR #4674, opened 2026-07-25
feat(ci): Remove Python version from build matrix

sgibson91:
### Checklist

- [x] I agree to follow _The Turing Way_'s [code of conduct](https://book.the-turing-way.org/community-handbook/coc/)
- [x] I have read the [contribution guide](https://book.the-turing-way.org/community-handbook/contributing/)
- [x] I have read the [style guide](https://book.the-turing-way.org/community-handbook/style/)

### Summary

This MR removes the Python version from the build matrix and passes it directly to the setup python step instead. We don't actually want to test multiple versions of Python, and bumping the Python version causes the branch protection rule to break since the matrix variables get put into the job names. The setup python step is the only place in this file that the python version is ingested.

- move python version out of build matrix and pass directly to setup python step instead

- [x] Everything looks ok?


.github/workflows/ci.yml
```diff
@@ -87,6 +87,19 @@ jobs:
           name: book-html-${{ github.run_id }}
           path: ./book/website/_build/html/
 
+  check-build-success:
+    # job to check the book has successfully built on all three OS's
+    runs-on: ubuntu-latest
+    needs: build-jb
+    if: always()
```
JimMadge: Is this needed? I would assume if you don't specify `if` a job always runs. Or, is `always()` not what I expect :eyes:

JimMadge (commented):
Oh, good catch.

So am I understanding correctly,

- Job names (or id) include the matrix element (?) like Python version and OS
- Required jobs are declared by name (or id) so whenever we change the matrix we also have to update the required jobs
- Which feels fragile, so another approach would be to check all jobs (up to a point) have run successfully

If using a matrix for Python is a problem, is there a way we can use a variable? Because we only actually test one Python version.

I think we only have it in the matrix because we don't want to have magic numbers/strings in the workflow.

.github/workflows/ci.yml
sgibson91: I think a few words might be missing from your comment. But basically this makes sure this job will run even if the build job, which this job depends on, fails. This will be the difference between the required job (the new one) showing up in the UI as failed or pending because it hasn't actually run.

sgibson91:
Yes, you've understood correctly.

We could just put the python version straight into the step that uses it on line https://github.com/sgibson91/the-turing-way/blob/3489c3e4189fe77d3feb2b45f45f2a492eaefd78/.github/workflows/ci.yml#L68 since it's not used anywhere else.

Without the job I'm adding though, we'd still have to pick one of the 3 OS's that we'd favour as being the "correct" test to pass. This will make it so all 3 are valued equally.

Moving the Python version out of the matrix does solve the problem of whether we need to update the branch protection rules every time we bump python. So I guess it depends on how much we care about a PR being mergeable if either the Mac or Windows builds break (we currently only require Linux to pass).

sgibson91:
Well, actually. I guess we can set all 3 matrix subjobs as required by the branch protection rule and not add this new job at all.

But yeah, if anything in the matrix changes to cause them to be named something different, you're essentially breaking branch protection.

.github/workflows/ci.yml
JimMadge: Yeah, reading that back it isn't very clear :laughing:.

I think what I was missing is that a job doesn't normally run if there have been failures :+1:.

JimMadge:
Oh, if `python_version` is only in one place it might make sense to remove it from the matrix.

IIRC, before we settled on only requiring the Ubuntu build because,

- It is most similar to the Netlify build
- It tends to be the quickest (and MacOS used to be _very_ slow)

I think for simplicity I would lean towards removing Python version from the matrix and keep the requirement for the Ubuntu build only.

.github/workflows/ci.yml
sgibson91: Ah yeah, now I understand your context, I understand the question! 😂

sgibson91:
Yeah, I'm happy to go with that, let me update this

sgibson91:
@JimMadge ok ready, I just need to update the branch protection rule once you've signed off

JimMadge (approved):

